PR #109, opened 2024-07-09
rename terms to tos

meetpateltech:
### Summary

Renames the '/terms' page to '/tos' to fix a 404 error.

### Details

- The '/terms' page was returning a 404 error because the terms of service page is actually located at '/tos'.
- Updates the link in the Sign In page from '/terms' to '/tos' to correctly point to the terms of service page.
- This is a minor change to fix a broken link and ensure the terms of service page is accessible to users.
  

<details>
The /terms page is giving a 404 error because the terms of service page is named /tos
</details>


Dhravya:
good catch! thanks

PR #111, opened 2024-07-09
show chunks seperations in database, proper error handling and prettier config

ThakerKush:
### Summary

This pull request introduces a new feature for storing and retrieving chunks of text from a vector database, along with improved error handling for processing large documents.

### Details

- **New Feature: Chunk ID's in Database:**
  - Added functionality to store chunk ID's in the database, allowing for efficient retrieval of specific chunks of text.
  - Implemented a new `batchCreateChunksAndEmbeddings` function in `helper.ts` to handle the creation and storage of chunks.
  - Modified the `index.ts` file to return the chunked input, which includes the chunk ID's and corresponding text.

- **Improved Error Handling:**
  - Enhanced error handling in `web/app/actions/doers.ts` to provide more informative error messages when processing large documents.
  - Added a check in `index.ts` to prevent the processing of documents with more than 20 chunks.

- **Refactoring:**
  - Introduced a `prettier` configuration file to enforce consistent code formatting.
  - Removed unnecessary imports and code in `web/app/actions/doers.ts`.
  - Updated the `createMemory` function in `web/app/actions/doers.ts` to handle the chunked input from the backend.

- **Other Changes:**
  - Updated the `systemPrompt` and `template` variables in `prompts/prompt1.ts` (likely related to the new chunking feature).

<details>
feat: improve error handling for long documents
feat: add chunk id's to db
</details>

### DIFF
--- a/.prettierrc
+++ b/.prettierrc
@@ -1,0 +1,5 @@
+{
+  "tabWidth": 2,
+  "trailingComma": "all",
+  "useTabs": false
+}
--- a/apps/cf-ai-backend/src/index.ts
+++ b/apps/cf-ai-backend/src/index.ts
@@ -15,7 +15,6 @@ import { zValidator } from "@hono/zod-validator";
 import chunkText from "./utils/chonker";
 import { systemPrompt, template } from "./prompts/prompt1";
 import { swaggerUI } from "@hono/swagger-ui";
-import { createOpenAI } from "@ai-sdk/openai";
 
 const app = new Hono<{ Bindings: Env }>();
 
@@ -65,14 +64,18 @@ app.post("/api/add", zValidator("json", vectorObj), async (c) => {
   const { store } = await initQuery(c);
 
   console.log(body.spaces);
-  await batchCreateChunksAndEmbeddings({
+  const chunks = chunkText(body.pageContent, 1536);
+  if (chunks.length > 20) {
+    return c.json({ status: "error", message: "no chunks over 20" });
+  }
+  const chunkedInput = await batchCreateChunksAndEmbeddings({
     store,
     body,
-    chunks: chunkText(body.pageContent, 1536),
+    chunks: chunks,
     context: c,
   });
 
-  return c.json({ status: "ok" });
+  return c.json({ status: "ok", chunkedInput });
 });
 
 app.post(
--- a/apps/cf-ai-backend/src/index.ts
+++ b/apps/cf-ai-backend/src/index.ts
@@ -66,7 +66,7 @@ app.post("/api/add", zValidator("json", vectorObj), async (c) => {
   console.log(body.spaces);
   const chunks = chunkText(body.pageContent, 1536);
   if (chunks.length > 20) {
-    return c.json({ status: "error", message: "no chunks over 20" });
+    return c.json({ status: "error", message: "We are unable to process documents this size just yet, try something smaller" });
   }
   const chunkedInput = await batchCreateChunksAndEmbeddings({
     store,
--- a/apps/web/app/actions/doers.ts
+++ b/apps/web/app/actions/doers.ts
@@ -7,7 +7,6 @@ import {
   chatHistory,
   chatThreads,
   contentToSpace,
-  sessions,
   space,
   spacesAccess,
   storedContent,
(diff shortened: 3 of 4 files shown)


apps/cf-ai-backend/src/index.ts
```diff
@@ -65,14 +64,18 @@ app.post("/api/add", zValidator("json", vectorObj), async (c) => {
   const { store } = await initQuery(c);
 
   console.log(body.spaces);
-  await batchCreateChunksAndEmbeddings({
+  const chunks = chunkText(body.pageContent, 1536);
+  if (chunks.length > 20) {
+    return c.json({ status: "error", message: "no chunks over 20" });
```
Dhravya: I assume this message is shown to the user, maybe we can keep the errors consistent for the user and just say "This document is too large" or something?

apps/web/app/actions/doers.ts
```diff
@@ -265,6 +262,40 @@ export const createMemory = async (input: {
...
+    status: string;
+    chunkedInput: string;
+    message?: string;
+  };
+
+  try {
+    if (response.status !== "ok") {
+      if (
+        response.status === "error" &&
+        response.message === "no chunks over 20"
```
Dhravya: matching the exact message can make it difficult to maintain, we can either:
1. use `data` as an numeric error code
2. prefix the message with 'ERROR:' and show the exact same error message to the user

apps/web/app/actions/doers.ts
```diff
@@ -349,32 +380,10 @@ export const createMemory = async (input: {
     );
   }
 
-  try {
```
Dhravya, line 352: not sure if removing this is a good idea, this existed to make sure that there's 'consistency' between the vector store and our document store. are we still making sure that the data is consistent? if yes, yeah removing it is a great idea

Dhravya (changes requested):

.prettierrc
```diff
@@ -0,0 +1,5 @@
+{
+  "tabWidth": 2,
+  "trailingComma": "all",
+  "useTabs": false
```
Dhravya, line 4: `useTabs` should be true, something I learnt recently

Dhravya (approved):

.prettierrc
Dhravya: .

Dhravya (changes requested):

Dhravya:
!feedback automatically changed title was not descriptive

apps/web/app/actions/doers.ts
ThakerKush: This had to be moved to account for error response form the /api/add endpoint ( the error related to the maximum chunk size ) and yeah the vector store and d1 store should be consistent 

But there is an edge case here and in main, if one fails and the other one succeeds then they won't be consistent........right?

apps/web/app/actions/doers.ts
ThakerKush: Yeah, gonna change this

apps/cf-ai-backend/src/index.ts
ThakerKush: How about I change this one to "We are unable to process documents this size just yet, try something smaller"?

apps/web/app/actions/doers.ts
Dhravya: Yeah, we gotta make sure that doesn't happen. If it does maybe we should call the delete endpoints to get rid of them?

ThakerKush: Yes that or like a exponential back off retry for the failed db call, if that fails to then call the delete endpoint?

apps/cf-ai-backend/src/index.ts
Dhravya: yes works!

apps/web/app/actions/doers.ts
Dhravya: Yes that would be absolutely amazing

Dhravya: we don't _need_ retries or backoffs, it's a good to have, as long as we just ensure both are the same

Dhravya (approved):

PR #116, opened 2024-07-12
ui changes :)

CodeTorso:


CodeTorso:
why the hell is cloudflare pages so slow, it gives me so much anxiety

PR #157, opened 2024-07-24
remove yarn

aryasaatvik:
# Remove Yarn package manager

## Overview
This pull request removes the `yarn` package manager dependency from the project. It addresses the incompatibility of `corepack` with `bun` as a package manager.

## Changes
- Key Changes:
    - Removed the `packageManager` field from both `apps/web/package.json` and the root `package.json`.
- New Features:
    - None
- Refactoring:
    - None

<details>
packageManager field in package.json is added by corepack. However, corepack doesn't support bun as a package manager
</details>

### DIFF
--- a/apps/web/package.json
+++ b/apps/web/package.json
@@ -2,7 +2,6 @@
 	"name": "@repo/web",
 	"version": "1.0.0",
 	"private": true,
-	"packageManager": "yarn@1.22.21",
 	"scripts": {
 		"dev": "next dev",
 		"build": "next build",
--- a/package.json
+++ b/package.json
@@ -35,7 +35,6 @@
 	"engines": {
 		"node": ">=18"
 	},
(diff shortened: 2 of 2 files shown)


Dhravya:
The reason this exists is because for some reason builds were failing on cf pages. 
this is also the reason that builds take a shit ton of time to happen. probably some kind of a dependency issue

well, it's a problem but one that I'm not too motivated to solve right now

aryasaatvik:
> The reason this exists is because for some reason builds were failing on cf pages. this is also the reason that builds take a shit ton of time to happen. probably some kind of a dependency issue
> 
> well, it's a problem but one that I'm not too motivated to solve right now

I closed it becuase I realized its needed for running turborepo scripts https://github.com/vercel/turbo/issues/4762. Updated the setup guide to include yarn as a prereq in #159

Issue #167, opened 2024-07-26
Clicking on a memory should open it in a New Tab


Welding-Torch:
If I click on this, it should not open on the same page as Supermemory (this leads to people navigating away from supmem)
![image](https://github.com/user-attachments/assets/8efe03cd-de7a-4846-ba19-be14ea81af1a)
Instead, clicking on it should open it in a new tab.

Dhravya:
good first issue

Dhravya Shah
https://dhravya.dev

dishit7:
on it

rishabhguptareal:
hey is this issue fixed as of now?

Welding-Torch:
Doesn't seem to be fixed, links are still opening in the same tab

rishabhguptareal:
can you assign this to me

rishabhguptareal:
I think this is solved already kindly check it once and then revert back for further processing

rishabhguptareal:
![Screenshot 2024-07-29 202734](https://github.com/user-attachments/assets/2e8c4287-e64b-4f20-abf1-1a1d490f1290)
target is already blank here so it is opening on new tab kindly check if am touching the right issue or issue is some different

Welding-Torch:
> kindly check if am touching the right issue or issue is some different

You're right, it is fixed

PR #169, opened 2024-07-26
Fix/clicking on a memory should open it in a new tab (#167)

dishit7:
# Fix: Open External Links in New Tab

## Overview
This pull request addresses the issue of external links opening in the same tab by adding `target="_blank"` and `rel="noopener noreferrer"` to the `Link` component for external sources.

## Changes
- Key Changes:
    - Added `target="_blank"` and `rel="noopener noreferrer"` to the `Link` component for external links.
- New Features:
    - None
- Refactoring:
    - None

<details>
This PR addresses the issue of external links opening in the same tab. I've added target="_blank" rel="noopener noreferrer" to the Link component for external sources. Due to setup limitations, I wasn't able to fully test this change. I would appreciate review and testing assistance from the maintainers or other contributors.
</details>


apps/cf-ai-backend/.gitignore
Dhravya, line 1: Wrangler.toml files are not supposed to be in gitignore! it's public information and still safe. however, the contributors shouldn't push their wrangler.toml files so that the deployment doesn't break

apps/cf-ai-backend/wrangler.toml
Dhravya, line 1: (that would bring this back)

apps/web/server/auth.ts
Dhravya, line 1: Revert this too pls!

package.json
Dhravya, line 1: nice. thanks for this

Dhravya (changes requested):
Thanks for the PR! Really appreciate it.

bring back wrangler.toml files, that's it :)

apps/web/server/auth.ts
dishit7: If I fix the wrangler.toml issue and make a commit that should fix this as well right cause I think the change is due to some blank space otherwise the file is unchanged.

CodeTorso:
I believe this is fixed

Issue #184, opened 2024-07-29
W key functions as a hotkey


fyzanshaik:
When I press W/w inside the add a note extension that pops up on any website it switches to some different tab or selects something else instead of just w character being pressed. 
Example any github repository I try to create a note, I press W and it clicks on the branch button available.

fyzanshaik:
This strange behavior can only be seen in github page. 
On any github repository page if you press the keys,
W : Opens branch information button
S : Opens repository search bar
J,K: Higlights top 2 file/folder name's  inside repository

PR #186, opened 2024-07-29
Fix: Disable GitHub shortcuts in text field

fyzanshaik:
**Problem:**  
On GitHub repository pages, certain keyboard shortcuts (`W`, `S`, `J`, `K`) interfere with the Supermemory.ai extension's text field. When users try to type notes, these keys trigger GitHub-specific actions (e.g., opening branch information, repository search bar, and highlighting files) instead of allowing text input.

**Fix:**  
Modified the event listener in the extension to prevent GitHub's keyboard shortcuts from triggering when the user is typing in the extension's text field. This ensures that typing in the text field works as expected without interference from GitHub's shortcuts.

**Applicability:**  
This fix specifically addresses the issue on GitHub repository pages where the conflict occurs. The fix does not impact other websites, as the problem is unique to GitHub's keyboard shortcuts.

### DIFF
--- a/apps/extension/content/ContentApp.tsx
+++ b/apps/extension/content/ContentApp.tsx
@@ -131,6 +131,13 @@ export default function ContentApp({
 				});
 			}
 		});
+		const handleKeyDown = (e: KeyboardEvent) => {
+			if (isPopoverOpen) {
+			  e.stopPropagation();
+			  e.preventDefault();
+			}
+		  };
+		  document.addEventListener('keydown', handleKeyDown, true);
 
 		const portalDiv = document.createElement("div");
 		portalDiv.id = "popover-portal";


fyzanshaik:
This fixes the issue #184 (My issue)

fyzanshaik:
I believe the line 
```
document.removeEventListener("mousemove", () => {});
```
in the cleanup function is unnecessary and potentially confusing. It's likely a leftover from previous code that wasn't properly cleaned up.I suggest we remove it.

apps/extension/content/ContentApp.tsx
```diff
@@ -139,8 +146,10 @@ export default function ContentApp({
 
 		return () => {
 			document.removeEventListener("mousemove", () => {});
+			document.removeEventListener('keydown', handleKeyDown, true);
+
 		};
-	}, []);
+	}, [isPopoverOpen]);
```
CodeTorso: Why should the code re run if the popover is opened or closed, I feel it's unnecessary

CodeTorso:
@Dhravya your thoughts ?

apps/extension/content/ContentApp.tsx
fyzanshaik: I see, fair that we won't need to run the code again and again. Will push the commit

Dhravya:
lgt.

PR #343, opened 2025-03-21
fix extension tweet import

CodeWithShreyans:
## Summary by CodeRabbit

- **Bug Fixes**
	- Enhanced content submission reliability by ensuring consistent type determination.
- **Refactor**
	- Simplified the process for classifying new content to eliminate redundant checks.


Dhravya:
Could this potentially break when the content is `undefined`? Can we add a specific check instead?

CodeWithShreyans:
From what I understand and from it's type, content can't be undefined

Dhravya (approved):

PR #370, opened 2025-08-17
better add-memory - lesser animations + lesser redundancy [finished]

CodeTorso:
this includes some of my work in add memory component

I want to combine link, file and note into one thing.

and also reduce the extra animations are just not good for a dialog that users will constantly open


MaheshtheDev:
thanks for contribution, but we are already working on improved version on this, so closing this PR

CodeTorso:
makes sense

PR #377, opened 2025-08-23
ui (memory detail): improved memory detail view and open chat

MaheshtheDev:
Before
image

After 
image


alexf37:
not sure i like the raw html. can we render the html instead if the content is html?

PR #469, opened 2025-10-09
fix: add button type to ModelSelector and ProjectSelector components

aryankeluskar:
### Current Issue
Clicking the model picker or project selector on the home screen would submit the chat form instead of opening the dropdown menu.

https://github.com/user-attachments/assets/0d69f133-2700-4d8b-8553-c8da95acf199

### Root Cause
Buttons inside a `<form>` element default to `type="submit"` in HTML. Without explicitly specifying `type="button"`, the ModelSelector and ProjectSelector buttons were being treated as submit buttons.

### Solution
Added `type="button"` to both selector buttons:
- `ModelSelector` component
- `ProjectSelector` component

This prevents form submission and allows the buttons to only execute their `onClick` handlers.

---

### DIFF
--- a/apps/web/components/model-selector.tsx
+++ b/apps/web/components/model-selector.tsx
@@ -48,6 +48,7 @@ export function ModelSelector({
 	return (
 		<div className="relative">
 			<Button
+				type="button"
 				variant="ghost"
 				className="flex items-center gap-1.5 px-2 py-1.5 rounded-md transition-colors"
 				onClick={() => !disabled && setIsOpen(!isOpen)}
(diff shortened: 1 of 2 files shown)


aryankeluskar:
@Dhravya Another issue bugging me is that the memory graph doesn't open when I am on any screen other than home page even though the button's always visible in menu. I can see that `Memories` component only renders on the home page, and that's why it doesn't load on settings/chat pages. Fixing this would require creating a separate component for graph modal which could be too big of a change for this PR. 

https://github.com/user-attachments/assets/84d582d1-2925-4620-ad2e-9f6788185a20

Dhravya:
I'm merging this - good catch

As for the graph being visible, we're working on a new version of the UI which will fix this issue 🫶

PR #473, opened 2025-10-09
fix: sync logout across tabs and handle 401 errors (closes #471)

### LINKED ISSUE #471: Session not invalidated across browser tabs after logout
When a user logs out from one browser tab, other open tabs remain active and continue making API requests with expired/invalid credentials. These requests receive 401 Unauthorized responses but keep retrying repeatedly, causing unnecessary backend 

Functionality everythign works (user is logged out correctly on the tab used), but session handling is inconsistent across tabs, leading to repeated unauthorized requests and potential performance issues.

https://github.com/user-attachments/assets/a421fe5e-7704-48aa-90c0-15f72fc707cf

iamrishu11:
### Problem:
When a user logs out from one browser tab, other open tabs remain active and continue making unauthorized API requests with expired tokens. These requests get 401 Unauthorized responses but keep retrying, causing unnecessary backend load.

### Solution:
- Implemented synchronization of the logout process across all tabs.
- When a user logs out, a logout event is broadcast to all other open tabs via `localStorage`. Each tab listens for this event and immediately redirects to the login page, stopping further unauthorized requests.
- 401 responses from fetch requests also trigger a logout event and broadcast to all tabs.
  
### Additional Details:
- The logout event is triggered using `localStorage.setItem` and listens for changes via the `storage` event.
- All tabs clear session state and redirect users to the login page immediately.
- This solution ensures that the app handles invalid credentials properly across all tabs and prevents unnecessary API calls.

### Testing:
- Tested the fix across multiple tabs to ensure that logout synchronization works as expected.
- Verified that 401 responses from API requests trigger the logout event and clear user sessions in all tabs.
  
Closes #471

### DIFF
--- a/apps/web/lib/cross-tab-logout.tsx
+++ b/apps/web/lib/cross-tab-logout.tsx
@@ -8,6 +8,7 @@ import { useRouter } from "next/navigation"
  *
  * Usage: Place <CrossTabLogoutHandler /> at the root of your app (e.g. in layout.tsx).
  */
+
 export function CrossTabLogoutHandler() {
   const router = useRouter()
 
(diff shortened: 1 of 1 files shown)


apps/web/lib/cross-tab-logout.tsx
```diff
@@ -0,0 +1,50 @@
...
+  }, [router])
+
+  useEffect(() => {
+    // Optionally, listen for 401 responses globally (if using fetch)
+    // You can patch window.fetch here to catch 401s and trigger logout
+    const origFetch = window.fetch
+    const customFetch = Object.assign(
+      async (input: RequestInfo | URL, init?: RequestInit): Promise<Response> => {
+        const response = await origFetch(input, init)
+        if (response.status === 401) {
```
jaswanth478: This **if statement** gonna  Logs out on every 401 — unsafe behavior (even in real 401 cases not related to auth).

 Probable Fix: restrict logout to auth APIs only (e.g., check request URL or headers).

iamrishu11: hey i just pushed the new changes can u review it now?, and thx for pointing it out!

MaheshtheDev:
this is not necessary and issue and pr is two different apps. but i appreciate your efforts. and in future PR's please make sure your code changes has proper comments ( what i mean by that is not too many comments and not too little). 

Once again thank you for your contributions

PR #491, opened 2025-10-16
fix: mount graph dialog globally to fix chat page issue

### LINKED ISSUE #490: [Bug] - Graph view modal not opening on chat pages
- **Environment:** Web app, all browsers
- **Steps to reproduce:**
  1. Navigate to any chat page (e.g., `/chat/123`)
  2. Click the graph view button in header 
  3. Modal doesn't open (but state is set as closing the chat opens the graph)

- **Expected:** Modal should open from any page
- **Actual:** Modal only opens when navigating back to home page

Image

vorahardik7:
## Description

Fixes #490 

### Problem
The graph view modal was only mounted in the `Memories` component (which was only rendering on homepage). When users clicked the "Graph View" button in the header while on a chat page (/chat/[id]), the Zustand state would update but no component was mounted to render the dialog, causing it to appear non-functional.

### Solution
Extracted the graph dialog into a separate `graph-dialog.tsx` component and mounted it in the navigation layout, making it available on all pages (similar to how Add Memory is handled).

### Demo

**Graph view now working on chat pages:**

supfix

#### 1. Created `apps/web/components/graph-dialog.tsx`
- Extracted the graph dialog logic from `Memories` component
- Contains all query logic, document injection, and dialog rendering
- Self-contained component with same functionality as before

#### 2. Modified `apps/web/components/memories.tsx`
- Removed duplicate graph dialog section
- Removed unused imports (`MemoryGraph`, `Dialog`, `useGraphModal`)
- Component now focuses only on memory list view

#### 3. Modified `apps/web/app/(navigation)/layout.tsx`
- Added `GraphDialog` import
- Mounted `<GraphDialog />` component globally
- Follows existing pattern used for `AddMemoryView`

#### Both `Memories` and `GraphDialog` use the same TanStack Query cache key: `["documents-with-memories", selectedProject]`

### DIFF
--- a/apps/web/components/graph-dialog.tsx
+++ b/apps/web/components/graph-dialog.tsx
@@ -1,4 +1,3 @@
-// apps/web/components/graph-dialog.tsx
 "use client"
 
 import { useAuth } from "@lib/auth-context"


MaheshtheDev (approved):
Great work and thanks for the contribution

PR #493, opened 2025-10-17
feat(tkg): opt‑in temporal view (asOf/window) + validity intervals; SDK/UI/docs, all behind a flag

MagellaX:
# PR description

Adds a small-but-mighty “Temporal view” so clients can query memories as of a moment or within a window, without changing default behavior. I introduce optional `validFrom`/`validUntil` on memory entries and beta params `asOf`, `validFromGte`, `validUntilLte`; thread them through TS/OpenAI/Python SDKs only when `enableTemporalQueries` is set; surface a tiny, flag-gated time control in the console that filters both the list and the graph (out-of-window nodes are gently dimmed, version chains stay intact); and document it as beta. Shipped dark via `NEXT_PUBLIC_ENABLE_TEMPORAL_QUERIES=false`. Server can ignore new params; rollback is literally flipping the flag. Minimal surface area today, clean runway for bitemporal/graph traversal tomorrow.

- **Why this works:** additive + flag-gated, no migrations, no default perf hit; aligns with existing relations/versioning for a smooth backend snap-in later.
- **Tests:** Zod parsing + invariants, SDK pass-through off/on, UI temporal predicate + snapshots; manual check shows identical UX with the flag off.
- **Rollout:** ship dark → internal → canary → public beta; instant revert via flag if needed.

Why this direction feels rock-solid (from a quick deep-dive): Look, we know that temporal knowledge graphs are the future of AI memory, think Zep's bitemporal magic turning static facts into evolving stories, but we don't need to boil the ocean to get there. i have kept this approach purely low-risk/high-reward: I'm just adding optional validity fields (validFrom/validUntil) and beta query params (asOf, time windows) that slot neatly onto the existing version chains and relations (updates/extends/derives) without touching defaults or requiring migrations. It's all feature-flagged, so today's users see zero change, and the server can ignore the new stuff until it's ready to snap in bitemporal indexing or invalidation logic. From the schemas to SDKs to the UI graph, if that sounds good, everything threads cleanly, minimal surface area, no perf traps, and a clear runway for the full TKG payoff (like tracking preference evolution without the "why did my agent forget my coffee order?" drama :) lol ). It's the engineering equivalent of laying tracks before the train arrives: harmless now, transformative later, and yeah, I did a little bit of resisting over-engineering it into a flux capacitor... iykyk what that means :)

### DIFF
--- a/apps/web/app/(navigation)/layout.tsx
+++ b/apps/web/app/(navigation)/layout.tsx
@@ -1,5 +1,6 @@
 "use client"
 
+import { GraphDialog } from "@/components/graph-dialog"
 import { Header } from "@/components/header"
 import { AddMemoryView } from "@/components/views/add-memory"
 import { useEffect, useState } from "react"
@@ -52,6 +53,8 @@ export default function NavigationLayout({
 					onClose={() => setShowAddMemoryView(false)}
 				/>
 			)}
+
+			<GraphDialog />
 		</div>
 	)
 }
--- a/apps/web/components/graph-dialog.tsx
+++ b/apps/web/components/graph-dialog.tsx
@@ -1,0 +1,243 @@
+"use client"
+
+import { useAuth } from "@lib/auth-context"
+import { $fetch } from "@repo/lib/api"
+import type { DocumentsWithMemoriesResponseSchema } from "@repo/validation/api"
+import { useInfiniteQuery } from "@tanstack/react-query"
+import { useCallback, useEffect, useMemo, useState } from "react"
+import type { z } from "zod"
+import { MemoryGraph } from "@repo/ui/memory-graph"
+import { Dialog, DialogContent } from "@repo/ui/components/dialog"
+import { ConnectAIModal } from "@/components/connect-ai-modal"
+import { AddMemoryView } from "@/components/views/add-memory"
+import { useChatOpen, useProject, useGraphModal } from "@/stores"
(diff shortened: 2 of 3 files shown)


MagellaX:
Any thoughts?? @Dhravya

MaheshtheDev:
@MagellaX this looks like making pure assumption that client execute will honor the temporal search though right?

MagellaX:
> @MagellaX this looks like making pure assumption that client execute will honor the temporal search though right?

Hey Mahesh, yep, you’re reading it right. Right now, the client is just wiring the optional params through so we have the contract ready; the servers still treat them as no‑ops. That’s why it’s all behind the NEXT_PUBLIC_ENABLE_TEMPORAL_QUERIES flag, and the docs call it out as beta. As soon as the backend starts honoring the filters, we’ll flip the flag and point the UI/tests at the real behaviour.

MagellaX:
So if you’re okay with that plan, you can probably merge this. The whole point of this was to provide a solid direction for this. Bringing up a foundational support will make sure that we can take things from there efficiently. Happy to follow up with anything else you need. @MaheshtheDev

MagellaX:
Also let me know about anything from your side, so we can get on a call to discuss what PRs ya'll are aiming for to bring, and also about potentially joining the team!

Dhravya:
Not planned as of now, but we will re-open this PR once we are done with the backend changes for the same.

PR #496, opened 2025-10-19
Fix excessive icon refetch made by importing bookmarks

fyzanshaik:
Twitter bookmark import was causing excessive icon refetching, making too many network requests during progress updates. If 1000 bookmakrs processed 1000 http requests are sent to fetch icon.
Screenshot From 2025-10-19 06-25-59

**Solution:**
Changed Twitter import button to use CSS background-image instead of `` elements to load the icon once and cache it.

**Changes:**
- Updated `createTwitterImportButton()` in `ui-components.ts` to use CSS background-image
- Added cleanup code in `updateTwitterImportUI()` to handle old `` elements
- Icon now loads once via `browser.runtime.getURL("/icon-16.png")` and stays cached

**Result:**
- Stops repeated icon requests during bookmark import updates
- Better performance, fewer API calls
- Same visual appearance, more efficient

**Note:** Same `` issue exists in ChatGPT, Claude, T3, and toast notifications. Should create a utility function to fix this pattern across all platforms.


MaheshtheDev:
@fyzanshaik i have tested in prod i don't see this issue at all both in browser network tab and extension network tab

FYI: the requests are going to just extension, not to any server. 
image

fyzanshaik:
@MaheshtheDev I verified it once again, and yes the requests are happening to the extension itself. I guess its not a big deal 

https://github.com/user-attachments/assets/242cbf44-6657-4efa-a4eb-51e6c3e1ca2a

PR #497, opened 2025-10-19
feat(chat): increase maxSteps to allow multiple tool-calling rounds

aryankeluskar:
I was facing this issue multiple times when it would do 2 tool calls (for either add+search or 2 search calls) and then abruptly stop. I brought this issue up with @sohamd22 and we thought it was Claude API that was down, but I looked into the code found this `maxSteps` set to 2 in `apps/web/components/views/chat/chat-messages.tsx` which would end the chat after the second tool-call result was fetched. 

For reference, I have attached a few images to this chat where I encountered this issue. 

IMG_4374
IMG_6743

And, I got the supermemory chat to work when I explicitly asked it to NOT add memories but rather just search.

IMG_2999

Therefore, I have increased it to 10 to support complex queries or just adding multiple memories per chat while also having the capacity to search. To be honest, 10 is an arbitrarily picked number but I think it's high enough for most people to not hit it.

### DIFF
--- a/apps/web/components/views/chat/chat-messages.tsx
+++ b/apps/web/components/views/chat/chat-messages.tsx
@@ -247,7 +247,7 @@ export function ChatMessages() {
 					},
 				},
 			}),
-			maxSteps: 10, // Allow multiple tool-calling rounds
+			maxSteps: 10,
 			onFinish: (result) => {
 				const activeId = activeChatIdRef.current
 				if (!activeId) return


MaheshtheDev (approved):
LGTM, thanks for your contribution @aryankeluskar

aryankeluskar:
Good to hear @MaheshtheDev

Dhravya (approved):

aryankeluskar:
Update: It works really well now! Thank you for merging these changes @MaheshtheDev 🤠

image

PR #498, opened 2025-10-19
docs: fixed naming convention of SDK usage examples

max-programming:
The Python code examples had the wrong class name. Changed that from `supermemory` to `Supermemory`

Python:
```diff
-client = supermemory()
+client = Supermemory()
```

The JavaScript/TypeScript code examples will work because it's a default import but still following the correct name of the class and to be consistent with the [NPM docs](https://www.npmjs.com/package/supermemory), I renamed it from `supermemory` to `Supermemory`

JS:
```diff
-import supermemory from 'supermemory';
+import Supermemory from 'supermemory';

-const client = new supermemory();
+const client = new Supermemory();
```

### DIFF
--- a/apps/docs/memory-api/sdks/native.mdx
+++ b/apps/docs/memory-api/sdks/native.mdx
@@ -31,7 +31,7 @@ pip install --pre supermemory
 import os
 from supermemory import Supermemory
 
-client = supermemory(
+client = Supermemory(
     api_key=os.environ.get("SUPERMEMORY_API_KEY"),  # This is the default and can be omitted
 )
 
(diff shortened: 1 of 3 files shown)


MaheshtheDev:
@max-programming thanks for your time for the contributions. but the name `supermemory` intentional

max-programming:
@MaheshtheDev It's not about the "name". The code in the docs will cause errors. Python will do, but TypeScript won't (because it's default import). But there's inconsistency between the NPM readme and the docs

I suggest you to check again the change I've made

From Claude Code's review:
> Python correctness: In Python, using lowercase supermemory() would have thrown a NameError since the class is defined as Supermemory. This fix corrects actual broken example code.

max-programming:
@MaheshtheDev I tested the code in both Python and TypeScript and my change works

How IDEs show the code from docs

TypeScript:
image

Python:
image

PR #517, opened 2025-10-24
feat: update app component to have a better loading screen

lirena00:
- Changed the logo image source in Heaedr to use a file based svg rather than url one.
- Enhanced the loading indicator with an animated SVG.
- Adjusted styles for better alignment and spacing in the loading section.
- Added the dark-transparent.svg file to the public directory.

The main reason for these changes was that the old loading screen felt pretty broken. The header rendering was inconsistent with the rest of the extension, and the loading text was invisible because it was using a default `text-white` class. On top of that, there was no min-height, which made the whole extension feel "jittery" when it first opened.

I also downloaded the `dark-transparent.svg` and added it locally. This way, the logo will still load even if the internet is down. (I mean, yeah, why would they use the extension without internet... but it's just more robust this way, you get the point.)

https://github.com/user-attachments/assets/b61cae9d-eca0-4ae4-8668-cd701a354b84


MaheshtheDev (approved):
thanks for contribution, looks good to me

PR #518, opened 2025-10-24
feat: add persistent memory retention UI in settings

aryankeluskar:
@sohamd22 and I were discussing how easy it to implement persistent retention of memories across user profiles, so supermemory doesn't have to fetch memories/documents with basic facts about the user everytime they send a message. I implemented this in a testing playground (demo video - https://github.com/aryankeluskar/persistent-user-context/blob/master/public/demo.mp4). Since we found this very easy to implement and high-value to display, I decided to add a UI in settings to show the static/dynamic memories for persistent user memory in the supermemory app as well. This PR does assume that the backend uses persistent memory retention using `/v4/profile`

These were the changes made to implement this feature:

1. `packages/validation/api.ts`
  - Added ProfileResponseSchema with full OpenAPI documentation
  - Includes static and dynamic fact arrays

2. `packages/lib/queries.ts`
  - Added useUserProfile hook following exact patterns from existing hooks
  - Uses containerTag (user email) for profile fetching

3. `apps/web/components/views/profile.tsx`
  - Added profile data fetching with session?.email as containerTag
  - Displays static and dynamic facts below billing section

Feel free to have a look and let me know if any further changes are required! The supermemory app chatbot is already miles ahead of most other chatbots I have used, and I hope that this feature would take it a step further.


aryankeluskar:
Just noticed that profile testing video isn't loading on GitHub so uploaded it on YouTube as well - https://youtu.be/63_ix6OSPDQ

aryankeluskar:
Update to this PR: I think so profiles are not yet implemented in the backend for chat? I could be wrong but when I ask the chatbot for its instructions and information about me, it didn't have access to the kind of information that `/v4/profiles` would have. I further tested this by simulating the same conversation across my user-profiles demo (https://youtu.be/63_ix6OSPDQ) and the chatbot on app.supermemory.ai, but it did not have the same static context in its system prompt.

image

MaheshtheDev:
i feel this requires more on the backend changes? instead of frontend IMO since chatbot needs to add profile. Will talk to team and get back to you

aryankeluskar:
Yeah, I agree. I don't think so the system prompt (and therefore the user context) is being prepared on the frontend, but these frontend changes should be applicable after it's implemented on backend as well.

MaheshtheDev:
@aryankeluskar can you add UI screenshot of these changes?

MaheshtheDev:
will reopen once we have those settings back up

PR #520, opened 2025-10-25
fix: resolve UI accessibility, hydration, and semantic HTML issues

amannhq:
- Add suppressHydrationWarning to body element to handle browser extension attributes
- Add DialogTitle components to graph-dialog and add-memory dialog for screen reader accessibility
- Convert SVG attributes to React/JSX camelCase format (strokeWidth, strokeLinecap, strokeLinejoin, clipPath)
- Replace div role="button" with semantic button element in chat sidebar
- Add biome-ignore comment for array index keys in text-effect animation (position is stable)
- Add localhost bypass in middleware to prevent authentication redirect loops during development
- Add lang="en" attribute to global-error.tsx html element
- Clean up duplicate CSS classes in add-memory component


apps/web/middleware.ts
```diff
@@ -7,6 +7,10 @@ export default async function middleware(request: Request) {
 	console.debug("[MIDDLEWARE] Path:", url.pathname)
 	console.debug("[MIDDLEWARE] Method:", request.method)
 
+	if (url.hostname === "localhost") {
```
MaheshtheDev: please remove this condition in middleware. when you running in the local you can have in middleware

amannhq: fixed

Issue #506, opened 2025-10-22
It's not working locally.


tzxylao:
mac M1 environment

node 20.19.5

Boot completed

Does this project have a backend service? Accessing localhost:8787 is not working, has the backend service been stripped away?

Stuck on the login page, clicking login reports Failed to fetch (localhost:8787)

Can you clarify my doubts?

MaheshtheDev:
@tzxylao , may i know from which root you are running the `bun run dev` and what is your env in `apps/web` folder?

tzxylao:
> [@tzxylao](https://github.com/tzxylao)* , may i know from which root you are running the `bun run dev` and what is your env in `apps/web` folder?

This is my configuration, with some parameters blurred.Is it missing any configuration? @MaheshtheDev 

```

NEXT_PUBLIC_BACKEND_URL=http://localhost:8787 

OPENAI_API_KEY=sk-xxx
  

PROVIDER_URL=https://ai.xxx.work/v1  
  

MODEL_NAME=gpt-5-mini-2025-08-07
  

DATABASE_URL=postgresql://root:xxxx@192.168.1.1:30406/supermemory
```

MaheshtheDev:
@tzxylao , value for backend_URL should be this

`NEXT_PUBLIC_BACKEND_URL=https://api.supermemory.ai`

tzxylao:
Does it mean that the backend service must rely on api.supermemory.ai? Can't the entire system be deployed independently? @MaheshtheDev

MaheshtheDev:
Memory engine is not `open-source` . but you will be able to run supermemory app locally with `api.supermemory.ai`

tzxylao:
Unable to log in

<img width="1045" height="487" alt="Image" src="https://github.com/user-attachments/assets/4d9e6b36-71b8-41fc-acb9-c932194d9cec" />

The email received the verification code, but after clicking it, I was still redirected back to the login page and did not see any obvious error messages.In addition to the aforementioned image in the lower left corner @MaheshtheDev

MaheshtheDev:
@tzxylao you can add some random `NEXT_PUBLIC_POSTHOG_KEY` env for now, but to be precise you can give your own posthog client key as well. However, in next update will make sure that it won't error out even when POSTHOG_KEY is not configured

tzxylao:
[...]

PR #526, opened 2025-10-26
Fix/invalid dom property

lirena00:
UI wise this doesn't change anything just addresses the issues #506 that were encountered while running it locally

* Added a descriptive `<DialogTitle>` ("Memory Dialog") to the add memory dialog in `AddMemoryView` for better accessibility and clarity. 

* Refactored the SVG markup in `ModelSelector` to streamline the structure, move attributes to the `<g>` element, and replace the `<rect>` in the clip path with a `<path>`, improving maintainability and consistency. (previously this was using raw svg code thus raising error converted it to jsx to fix the issue)

### DIFF
--- a/apps/web/components/views/add-memory/index.tsx
+++ b/apps/web/components/views/add-memory/index.tsx
@@ -543,7 +543,7 @@ export function AddMemoryView({
 					}}
 					open={showAddDialog}
 				>
-					<DialogTitle>Memory Dialog</DialogTitle>
+					<DialogTitle className="sr-only">Memory Dialog</DialogTitle>
 					<DialogContent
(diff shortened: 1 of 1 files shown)


apps/web/components/model-selector.tsx
```diff
@@ -55,41 +55,26 @@ export function ModelSelector({
 				disabled={disabled}
 			>
 				<svg
-					xmlns="http://www.w3.org/2000/svg"
-					width="24"
-					height="24"
-					viewBox="0 0 24 24"
-					fill="none"
+				xmlns="http://www.w3.org/2000/svg"
```
MaheshtheDev, line 58: may i know why you changed this `svg`

lirena00: oh I didn't
it's just I used https://svg2jsx.com/ to convert svg to jsx syntax and it changed it to that 
it still looks the same 
image

PR #535, opened 2025-10-28
Refactor browser extension content scripts to use platform adapters

MagellaX:
## Summary
- introduce a tiny PlatformAdapter interface so each host (ChatGPT, Claude, etc.) can plug in cleanly
- move ChatGPT and Claude scripts onto the adapter pattern and have the loader iterate adapters instead of hard-coded checks
- lay the groundwork for adding new chat hosts without duplicating the DOM-wiring logic


MaheshtheDev:
@MagellaX , we can add for T3 as well right?

MagellaX:
> @MagellaX , we can add for T3 as well right?

Hey yeah!, yep:-T3 is on my radar. Once the adapter refactor lands I’ll apply the same pattern there and circle back with the update.

PR #558, opened 2025-11-04
fix: add integer and minimum validation for limit parameter

karamvirsingh1998:
HEY @MaheshtheDev @Dhravya 
I was trying to integrate the api to lsit the documents , i accidently put negative number and it returned , there i catched one bug , so  i tried to test it on rest of the things also

I found this
1. It is not working on decimal numbers [ Returning on server error]
2. It is returning even on negative numbers

I have added fixes for it and tested it . Please review and Merge
image
image
image

### DIFF
--- a/packages/validation/api.ts
+++ b/packages/validation/api.ts
@@ -269,6 +269,12 @@ export const ListMemoriesQuerySchema = z
 			.regex(/^\d+$/)
 			.or(z.number())
 			.transform(Number)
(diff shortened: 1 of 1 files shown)


karamvirsingh1998:
@MaheshtheDev  Can you review this

MaheshtheDev:
thanks for your suggestion but these changes doesn't certainly effect on backend. will make sure we do on the backend though

PR #560, opened 2025-11-05
Migrate Chat Persistence from localStorage to IndexedDB to Fix QuotaExceededError

aryankeluskar:
image

I encountered a `QuotaExceededError` when I sent a new message, which was because. This occured because the chat state (managed via Zustand's `persist` middleware in `stores/chat.ts`) is stored in localStorage under the key `'supermemory-chats'`. Since my chat history containing messages and tool outputs was large enough, I got this error.

Overall, this limits scalability of the current chat persistence users with extensive chat data and it can lead to loss of chat history if they delete the localStorage to make room for new chats.

To resolve this, I migrated the persistence layer from localStorage to IndexedDB, which offers significantly higher storage limits (as described in this [stackexchange post](https://softwareengineering.stackexchange.com/questions/219953/how-is-localstorage-different-from-indexeddb)) and it has better performance for larger datasets.

I refactored the code to use the lightweight `idb-keyval` library as an IndexedDB wrapper, which integrates directly with Zustand's `createJSONStorage`. This adapter also includes automatic migration logic to transfer any existing data from localStorage to IndexedDB on first access.


MaheshtheDev (changes requested):
@aryankeluskar can you please resolve merge conflicts?

aryankeluskar:
@MaheshtheDev should be fixed now!

PR #599, opened 2025-12-01
feat(tools): allow passing apiKey via options for browser support

codewarnab:
## Problem
withSupermemory only read the API key from process.env.SUPERMEMORY_API_KEY, which does not work in browser environments.

## Solution
- Added optional apiKey field to WithSupermemoryOptions
- Falls back to process.env.SUPERMEMORY_API_KEY if not provided
- Updated middleware to accept and use the passed API key
- Updated README documentation

## Usage
Browser usage - pass apiKey directly:
const model = withSupermemory(openai('gpt-4'), 'user-123', { apiKey: 'your-key-here' });

Server usage - still works with env variable:
const model = withSupermemory(openai('gpt-4'), 'user-123');"


MaheshtheDev (approved):
tweaked few changes and updated the documentations. Now it all looks goo

PR #625, opened 2025-12-22
feat: Add advanced tools integration guide for llm-advanced-tools

hetpatel-11:
- Add documentation showing how to integrate llm-advanced-tools with Supermemory tools
- Demonstrate 85%+ token savings via tool search and defer loading (According to Anthropic)
- Include complete example code for users
- Add example TypeScript file showing integration pattern

This allows users to reduce context waste and improve performance when using Supermemory tools with many other tools. Anthropic only supports claude models that is why I created advanced tools to make it model agnostic, use it with any LLM in the Vercel AI SDK. 
Here is my repo: https://github.com/hetpatel-11/advanced-tools

I do not have access to your tools since you wrap them through your function, so you will need to add this internally. 
image

### DIFF
--- a/apps/docs/ai-sdk/advanced-tools-integration.mdx
+++ b/apps/docs/ai-sdk/advanced-tools-integration.mdx
@@ -1,0 +1,252 @@
+---
+title: "Advanced Tools Integration"
+description: "Reduce context waste and improve performance with llm-advanced-tools integration"
+sidebarTitle: "Advanced Tools"
+---
+
+<Note>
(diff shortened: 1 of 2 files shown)


Dhravya:
Sorry, we will not be able to merge this at the time. But the idea is very promising and we will continue to look into this!

hetpatel-11:
Hey @Dhravya , no need to merge it. Just give it a try, it should remove context upfront and improve performance. I just wanted you to try it and if it helps then add it :).

PR #627, opened 2025-12-23
MemoryGraph - revamped

vrupak:
# Memory Graph Changelog

## Development Setup

To test changes, run these commands in separate terminals:

**Terminal 1** - Install the required dependencies:
```bash
bun install
```

**Terminal 1** - Build memory-graph in watch mode:
```bash
cd packages/memory-graph && bun run dev
```

**Terminal 2** - Run the playground:
```bash
cd apps/memory-graph-playground && bun run dev
```

Then open http://localhost:3000 in your browser.

---

#### Slideshow Mode
Auto-cycling through nodes with smooth animations and physics simulation
- Random node selection every 3.5s (avoids consecutive duplicates)
- Smooth pan-to-node animation with automatic popover
- Brief physics pulse (1s) on each selection
- Background dimming animation
- Single-click to stop

#### Node Popover with Background Dimming
Floating popover with smart positioning and focus dimming effect
- Smooth 1.5s cubic ease-out dimming animation
- Non-selected nodes: 20% opacity, unconnected edges: 10% opacity
- Smart edge detection with 20px gap from node
- Auto-flips to avoid viewport edges
- Close via backdrop click, X button, or Escape key
- Shows: title, summary, type, memory count, URL, date, ID

#### Document Type Icons
Canvas-rendered icons centered on document cards
- Supported: TXT, PDF, MD, DOC/DOCX, RTF, CSV, JSON
- Scales with card size (40% of height)
- Only renders when zoomed in

#### Physics-Driven Layout
Simplified initial positioning, letting physics create natural layouts
- Simple grid with random offsets (no concentric rings)
- 50 quick pre-ticks + smooth animation
- Eliminates teleportation on node interaction
- Faster, non-blocking initial render

#### Updated Color Scheme
Refined palette for better contrast and readability

#### Edge Viewport Culling
Fixed edges disappearing during zoom/pan
- Now checks both X and Y axis bounds
- Only culls when BOTH endpoints off-screen in same direction
- 100px margin on all sides

#### Memory Nodes Follow Parents
Memory nodes now move with parent documents when dragged
- Store relative offset instead of absolute position
- Automatically repositions based on parent location

#### k-NN Similarity Algorithm
Reduced from O(n²) to O(n·k)
- 3x faster: ~50ms → ~17ms for 100 docs
- 4,950 → 1,500 comparisons for 100 docs
- Separated into own memo (doesn't recalculate on UI interactions)

#### Memory Leak Fix
NodeCache now cleans up deleted nodes properly

#### Race Condition Fix
Atomic node/edge updates eliminate NaN positions

#### Canvas Rendering Optimizations
Reduced per-frame overhead and improved rendering efficiency
- Spatial grid for hit detection
- Batched edge rendering by type (fewer canvas state changes)
- Canvas quality settings initialized once instead of every frame
- Optimized render key using fast hash instead of string concatenation
- Memoized nodeMap to avoid rebuilding every frame

#### Node Limiting & Memory Management
Smart memory limiting prevents performance issues with large datasets
- `maxNodes` prop limits total memory nodes (default: 500 in playground)
- Dynamic per-document cap distributes budget across documents
- Prioritizes recent memories and high-relevance scores
- k-NN similarity limit reduced from 15 to 10 connections per document

---

### DIFF
--- a/apps/docs/connectors/github.mdx
+++ b/apps/docs/connectors/github.mdx
@@ -410,14 +410,27 @@ When you update the repository configuration:
   <Tab title="TypeScript">
     ```typescript
     // Delete by connection ID
-    const result = await client.connections.delete(connectionId);
+    const result = await client.connections.deleteByID(connectionId);
+
+    // Or delete by provider (requires container tags)
+    const result = await client.connections.deleteByProvider('github', {
+      containerTags: ['user-123']
+    });
+
     console.log('Deleted connection:', result.id);
     ```
   </Tab>
   <Tab title="Python">
     ```python
     # Delete by connection ID
-    result = client.connections.delete(connection_id)
+    result = client.connections.delete_by_id(connection_id)
+
+    # Or delete by provider (requires container tags)
+    result = client.connections.delete_by_provider(
+        provider='github',
+        container_tags=['user-123']
+    )
+    
     print(f'Deleted connection: {result.id}')
     ```
   </Tab>
--- a/apps/docs/connectors/overview.mdx
+++ b/apps/docs/connectors/overview.mdx
@@ -288,11 +288,16 @@ const client = new Supermemory({
   apiKey: process.env.SUPERMEMORY_API_KEY!
 });
 
-// Delete by connection ID using SDK
-const result = await client.connections.delete(connectionId);
+// Delete by connection ID
+const result = await client.connections.deleteByID(connectionId);
+
+// Or delete by provider (requires container tags)
+const result = await client.connections.deleteByProvider('notion', {
+  containerTags: ['user-123']
+});
 
 console.log('Deleted:', result.id, result.provider);
-// Output: Deleted: conn_abc123 notion
+
 ```
 
 ```python Python
@@ -301,11 +306,17 @@ import os
 
 client = Supermemory(api_key=os.environ.get("SUPERMEMORY_API_KEY"))
 
-# Delete by connection ID using SDK
-result = client.connections.delete(connection_id)
+# Delete by connection ID
+result = client.connections.delete_by_id(connection_id)
+
+# Or delete by provider (requires container tags)
+result = client.connections.delete_by_provider(
+    provider='notion',
+    container_tags=['user-123']
+)
 
 print(f"Deleted: {result.id} {result.provider}")
-# Output: Deleted: conn_abc123 notion
+
 ```
 
 ```bash cURL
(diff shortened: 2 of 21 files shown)


nexxeln (changes requested):
Image

getting this error on the playground

```
## Error Type
Runtime InvalidStateError

## Error Message
CanvasRenderingContext2D.scale: Canvas exceeds max size.


    at zn</< (../../packages/memory-graph/src/components/graph-canvas.tsx:967:9)
    at ui (../../packages/memory-graph/src/components/memory-graph.tsx:728:6)
    at Home (src/app/page.tsx:274:7)

## Code Frame
  965 |
  966 | 			const ctx = canvas.getContext("2d")
> 967 | 			ctx?.scale(dpr, dpr)
      | 			     ^
  968 | 		}, [width, height, dpr])
  969 | 		// -----------------------------------------------------------------------
  970 |

Next.js version: 16.0.3 (Turbopack)
```

vrupak:
@nexxeln check now?

nexxeln:
the width seems to be broken
image

when i zoomed out i could see the graph. but then it started flickering and vanished
image

vrupak:
there were 2 potential issues which i addressed
1. the canvas element still had height={height} as a prop, but since we are setting different dimensions inside the useLayoutEffect. it probably caused conflicts.
2. when width or height is 0 or very small, MAX_CANVAS_SIZE / width could be infinity, which would break the calculation. 

lmk if its still not working

packages/memory-graph/src/components/memory-graph.tsx
```diff
@@ -126,6 +133,31 @@ export const MemoryGraph = ({
 		nodePositions,
 		draggingNodeId,
 		memoryLimit,
+		maxNodes,
+	)
+
+	// State to trigger re-renders when simulation ticks
+	const [, setSimulationTick] = useState(0)
```
nexxeln: i think the proper way to do this would be something like a useReducer

```typescript
const [, forceRender] = useReducer((x: number) => x + 1, 0)
```

and in the hook

```typescript
export function useForceSimulation(
  nodes: GraphNode[],
  edges: GraphEdge[],
  onSimulationTick: () => void,
  enabled = true,
): ForceSimulationControls {
  // ...
  .on("tick", onSimulationTick)
```

nexxeln (changes requested):
pls change this weird pattern also

vrupak:
@nexxeln done with the changes.

I will be travelling both today and tomo, so wont be as active for a while. Also i have shared the repo access with you, if you want you can contribute as well.

PR #634, opened 2025-12-29
fix(tools): pass apiKey to profile search instead of using process.env

codewarnab:
## Summary

Fixes a regression where the API key passed via `options.apiKey` in `withSupermemory()` was not being used for profile search requests. Instead, the code was hardcoded to use `process.env.SUPERMEMORY_API_KEY`, causing failures in environments where the environment variable is not set (for example, Vercel middleware).

## Root Cause

This regression was introduced in PR #628 (`feat(@supermemory/tools): vercel ai sdk compatible with v5 and v6`). That change rewrote `memory-prompt.ts` but did not preserve the `apiKey` parameter flow that was originally added in PR #599 (`feat(tools): allow passing apiKey via options for browser support`).

## Changes

- Added an `apiKey` parameter to the `supermemoryProfileSearch` function  
- Added an `apiKey` parameter to the `addSystemPrompt` function  
- Updated the `Authorization` header to use the passed `apiKey` instead of `process.env.SUPERMEMORY_API_KEY`  
- Updated `transformParamsWithMemory` in `middleware.ts` to pass `ctx.apiKey` to `addSystemPrompt`

## Behavior

The API key resolution priority remains:

1. `options.apiKey` (if provided) — highest priority  
2. `process.env.SUPERMEMORY_API_KEY` — fallback  

This is resolved in `index.ts` (line 71):

```ts
const providedApiKey = options?.apiKey ?? process.env.SUPERMEMORY_API_KEY;
````

This fix ensures that the resolved `apiKey` is consistently propagated to all API calls, including profile search requests.

### DIFF
--- a/packages/tools/src/vercel/memory-prompt.ts
+++ b/packages/tools/src/vercel/memory-prompt.ts
@@ -16,22 +16,23 @@ const supermemoryProfileSearch = async (
 	containerTag: string,
 	queryText: string,
 	baseUrl: string,
+	apiKey: string,
 ): Promise<ProfileStructure> => {
 	const payload = queryText
 		? JSON.stringify({
-				q: queryText,
-				containerTag: containerTag,
-			})
+			q: queryText,
+			containerTag: containerTag,
+		})
 		: JSON.stringify({
-				containerTag: containerTag,
-			})
+			containerTag: containerTag,
+		})
 
 	try {
 		const response = await fetch(`${baseUrl}/v4/profile`, {
 			method: "POST",
 			headers: {
 				"Content-Type": "application/json",
-				Authorization: `Bearer ${process.env.SUPERMEMORY_API_KEY}`,
+				Authorization: `Bearer ${apiKey}`,
 			},
 			body: payload,
 		})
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: packages/tools/src/vercel/memory-prompt.ts [hunk 1]
```diff
@@ -16,22 +16,23 @@ const supermemoryProfileSearch = async (
...
+			containerTag: containerTag,
+		})
 
 	try {
 		const response = await fetch(`${baseUrl}/v4/profile`, {
 			method: "POST",
 			headers: {
 				"Content-Type": "application/json",
-				Authorization: `Bearer ${process.env.SUPERMEMORY_API_KEY}`,
+				Authorization: `Bearer ${apiKey}`,
```
Dhravya, line 35: For backwards compatibility, we should probably still do process.env.SUPERMEMORY_API_KEY ?? apiKey

codewarnab: Backward compatibility is already maintained here:
https://github.com/supermemoryai/supermemory/blob/main/packages/tools/src/vercel/index.ts#L71

However, it gives higher priority to `options.apiKey` if set. I can change this if you prefer otherwise.

MaheshtheDev (approved):
good catch, @codewarnab . thanks for your contributions

packages/tools/src/vercel/memory-prompt.ts
MaheshtheDev: that make sense

PR #709, opened 2026-01-26
fix: await subscription checks

SyedaAnshrahGillani:
- Problem: fetchSubscriptionStatus never awaited the response from Autumn. check, so every plan was labeled as not allowed and missing status, even when data was returned. This meant the UI always treated users as having no active product.
 
  - Fix: Await each autumn. Check call before reading res.data, ensuring the fetched subscription status actually reflects the backend response.

### DIFF
--- a/packages/lib/queries.ts
+++ b/packages/lib/queries.ts
@@ -23,7 +23,7 @@ export const fetchSubscriptionStatus = (
 			await Promise.all(
 				allPlans.map(async (plan) => {
 					try {
-						const res = autumn.check({
+						const res = await autumn.check({
 							productId: plan,
 						})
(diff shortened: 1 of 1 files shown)


packages/lib/queries.ts
```diff
@@ -23,7 +23,7 @@ export const fetchSubscriptionStatus = (
 			await Promise.all(
 				allPlans.map(async (plan) => {
 					try {
-						const res = autumn.check({
+						const res = await autumn.check({
```
MaheshtheDev, line 26: autumn.check() from autumn-js/react is synchronous (not a Promise)

MaheshtheDev:
autumn.check() from autumn-js/react is synchronous (not a Promise)

PR #733, opened 2026-02-12
fix(mcp): register handler on bare /mcp path to match MCP transport spec

nmindz:
## Problem

The Supermemory MCP server returns `404 Not Found` when MCP clients send requests to the bare `/mcp` endpoint. This breaks compatibility with standard MCP clients such as [mcp-cli](https://github.com/philschmid/mcp-cli) and any client that follows the MCP Streamable HTTP transport specification.

The root cause is the Hono route pattern `app.all("/mcp/*")` in `apps/mcp/src/index.ts`. In Hono, the `/*` wildcard requires at least one path segment after `/mcp/` — meaning `POST /mcp` does not match, while `POST /mcp/sse` or `POST /mcp/message` would.

When mcp-cli connects using the standard HTTP transport configuration:

```json
{
  "type": "http",
  "url": "https://mcp.supermemory.ai/mcp",
  "headers": { "Authorization": "Bearer <token>" }
}
```

It sends `POST /mcp` with the JSON-RPC `initialize` request. The server responds with 404 before the request ever reaches the auth middleware or the MCP handler.

For comparison, the [NeonDB MCP server](https://github.com/neondatabase/mcp-server-neon) (`mcp-server-neon`) handles the same `POST /mcp` request correctly because its `mcp-handler` library matches exact paths rather than relying on wildcard patterns. Running `mcp-cli -d neon` succeeds while `mcp-cli -d supermemory-mcp` fails with:

```
Error [SERVER_CONNECTION_FAILED]: Failed to connect to server "supermemory-mcp"
Details: Streamable HTTP error: Error POSTing to endpoint: Not Found
```

## What the MCP specification requires

The [MCP Streamable HTTP transport specification](https://modelcontextprotocol.io/specification/2025-03-26/basic/transports) (revision 2025-03-26) states:

> An MCP server that supports Streamable HTTP transport MUST provide a single HTTP endpoint path (hereafter referred to as the MCP endpoint) that accepts requests.

The server MUST accept `POST` (client-to-server JSON-RPC messages), `GET` (opening an SSE stream for server-to-client notifications), and `DELETE` (session termination) on that single endpoint. The client sends requests directly to the configured URL — in this case `/mcp` — not to a sub-path of it.

Full specification reference: https://modelcontextprotocol.info/

The mcp-cli implementation (https://github.com/philschmid/mcp-cli) follows this spec correctly: it POSTs the JSON-RPC `initialize` request directly to the configured URL, without appending any sub-path. Any MCP server that does not respond to requests on the exact configured endpoint path is non-compliant.

### `apps/mcp/src/index.ts`

**1. Added `Context` type import from Hono (line 2)**

Required for the type annotation on the extracted handler function.

```typescript
// Before
import { Hono } from "hono"

// After
import { Hono, type Context } from "hono"
```

**2. Extracted inline route handler to a named `handleMcpRequest` function (lines 103-175)**

The authentication logic and Durable Object delegation are completely unchanged. The handler body was moved out of the inline `app.all()` callback into a named const to avoid duplicating it across two route registrations.

**3. Registered the handler on both `/mcp` and `/mcp/*` (lines 177-178)**

```typescript
// Before
app.all("/mcp/*", async (c) => {
  // auth + handler logic inline
})

// After
const handleMcpRequest = async (c: Context<{ Bindings: Bindings }>) => {
  // same auth + handler logic, unchanged
}

app.all("/mcp", handleMcpRequest)
app.all("/mcp/*", handleMcpRequest)
```

The bare `/mcp` registration handles the standard MCP client behavior (POST/GET/DELETE to the exact endpoint path). The `/mcp/*` wildcard registration preserves backward compatibility with any clients or internal routing that may use sub-paths.

No logic changes. No new dependencies. No changes to authentication, tool definitions, resources, or prompts.

## How other HTTP MCP servers handle this correctly

**NeonDB MCP server** (`mcp-server-neon`): Uses the `mcp-handler` library with `createMcpHandler()`, which internally matches exact paths. Their route handler in `landing/app/api/[transport]/route.ts` also includes explicit path normalization (lines 685-703) that rewrites `/mcp` to `/api/mcp` before passing to the handler, ensuring the endpoint always matches regardless of how the client addresses it:

```typescript
const handleRequest = (req: Request) => {
  const url = new URL(req.url);
  if (url.pathname === '/mcp') {
    url.pathname = '/api/mcp';
  }
  // ...
  return authHandler(normalizedReq);
};
```

This is the correct pattern: the server must be reachable at the exact path the client is configured to use, per the MCP Streamable HTTP transport specification.

**mcp-cli** (https://github.com/philschmid/mcp-cli): The reference client implementation POSTs directly to the configured URL without modification. When a server's routing does not match that exact path, the connection fails. This is the expected and spec-compliant client behavior.

## Testing

After this fix:

- `mcp-cli -d supermemory-mcp` successfully connects, completes the MCP initialize handshake, and lists available tools
- Existing clients using sub-paths under `/mcp/*` continue to work without changes
- Authentication behavior is unchanged: 401 without a token, 401 with an invalid token, normal flow with a valid token

### DIFF
--- a/apps/mcp/src/index.ts
+++ b/apps/mcp/src/index.ts
@@ -100,9 +100,6 @@ const mcpHandler = SupermemoryMCP.mount("/mcp", {
 	},
 })
 
-// MCP request handler — registered on both bare and wildcard paths
-// to comply with MCP Streamable HTTP transport spec, which requires
-// a single endpoint that accepts POST, GET, and DELETE.
 const handleMcpRequest = async (c: Context<{ Bindings: Bindings }>) => {
 	const authHeader = c.req.header("Authorization")
 	const token = authHeader?.replace(/^Bearer\s+/i, "")
--- a/apps/mcp/src/index.ts
+++ b/apps/mcp/src/index.ts
@@ -6,17 +6,17 @@ import { initPosthog } from "./posthog"
 import type { ContentfulStatusCode } from "hono/utils/http-status"
 
 type Bindings = {
-  MCP_SERVER: DurableObjectNamespace
-  API_URL?: string
-  POSTHOG_API_KEY?: string
+	MCP_SERVER: DurableObjectNamespace
+	API_URL?: string
+	POSTHOG_API_KEY?: string
 }
 
 type Props = {
-  userId: string
-  apiKey: string
-  containerTag?: string
-  email?: string
-  name?: string
+	userId: string
+	apiKey: string
+	containerTag?: string
+	email?: string
+	name?: string
 }
 
 const app = new Hono<{ Bindings: Bindings }>()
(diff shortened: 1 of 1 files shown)


Prasanna721:
please run biome ci

apps/mcp/src/index.ts
```diff
@@ -25,151 +25,157 @@ const DEFAULT_API_URL = "https://api.supermemory.ai"
...
-			email: authUser.email,
-			name: authUser.name,
-		} satisfies Props,
-	} as ExecutionContext & { props: Props }
-
-	return mcpHandler.fetch(c.req.raw, c.env, ctx)
-})
+// MCP request handler — registered on both bare and wildcard paths
+// to comply with MCP Streamable HTTP transport spec, which requires
+// a single endpoint that accepts POST, GET, and DELETE.
```
Prasanna721: remove the comments

Prasanna721:
Looks good, can you just remove the comments?

Prasanna721 (approved):

apps/mcp/src/index.ts
nmindz: I already did. Your changes referenced commit [f22900d](https://github.com/supermemoryai/supermemory/pull/733/commits/f22900d4078866997d2e074959c62f12c66fb21f), which is 2 commits behind the tip of the PR branch: [93911d6](https://github.com/supermemoryai/supermemory/pull/733/commits/93911d683850b18ac44f5abe61654ab656bf5fc3) (the latest commit message is literally: `remove extra comments`)

nmindz:
Any luck on getting this merged?

MaheshtheDev (approved):
LGTM

PR #750, opened 2026-02-20
Add 'type' field to supermemory MCP server config

robertmclaws:
Added 'type' field to supermemory MCP server configuration. Supermemory uses SSE, and if that is not directly specified, the MCP server will error out.


robertmclaws:
@Dhravya Did you get a chance to look into whether the MCP server is actually SSE or not? Claude Code's analysis says it is.

MaheshtheDev:
this has been updated to HTTP streamable, thanks for the PR though

PR #793, opened 2026-03-21
todo: Implement account deletion API call

ishaanxgupta:
Implemented the account deletion API call within the Account settings component, replacing a stubbed console log with a functional call to authClient.deleteUser().

### DIFF
--- a/apps/web/components/settings/account.tsx
+++ b/apps/web/components/settings/account.tsx
@@ -13,6 +13,8 @@ import {
 	DialogClose,
 } from "@ui/components/dialog"
 import { authClient } from "@lib/auth"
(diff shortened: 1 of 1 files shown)


ishaanxgupta:
@MaheshtheDev please have a look

MaheshtheDev (commented):
Thanks for the PR ishaan, its bit more complicated then just deleting the user. Cause supermemory account runs by organisation. so we have to handle that as well

ishaanxgupta:
> Thanks for the PR ishaan, its bit more complicated then just deleting the user. Cause supermemory account runs by organisation. so we have to handle that as well

Okay got it, should I close this one then?

PR #813, opened 2026-03-27
fix: handle invalid JSON response in facets query

cyphercodes:
## Problem
The dashboard shows a JSON parse error on 'Total memories stored' display (Issue #802).

Error: `Unexpected non-whitespace character after JSON at position 4`

This happens when the facets API endpoint returns unexpected data that isn't valid JSON, causing the frontend to fail when trying to parse the response.

## Solution
Add validation for the facets API response to ensure it has the expected shape before returning it. If the response is invalid, fall back to empty facets (`[]`) and zero total (`0`).

## Changes
- Added runtime validation for facets response data in `memories-grid.tsx`
- Returns safe default values ({ facets: [], total: 0 }) when response is invalid
- Prevents JSON parse errors from breaking the UI

## Testing
- Verified the fix handles malformed responses gracefully
- UI continues to work with fallback values

Fixes #802

### DIFF
--- a/apps/web/components/memories-grid.tsx
+++ b/apps/web/components/memories-grid.tsx
@@ -156,7 +156,21 @@ export function MemoriesGrid({
 				throw new Error(response.error?.message || "Failed to fetch facets")
 			}
 
-			return response.data as FacetsResponse
+			// Validate response data to prevent JSON parse errors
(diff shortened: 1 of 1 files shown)


MaheshtheDev:
don't need this facet here

PR #824, opened 2026-04-03
perf: Execute Claude memory tool calls concurrently

ishaanxgupta:
This PR refactores the` realClaudeMemoryExample` and `processClaudeResponse` functions inside `packages/tools/test/claude-memory-real-example.ts` to use Promise.all() to process incoming tool calls in parallel instead of sequentially via a for...of loop.

Claude API often sends multiple independent tool calls (like reading multiple distinct paths from memory in a single block). Iterating through and awaiting them individually causes an N+1 problem.


MaheshtheDev (approved):
thanks for the contribution

PR #830, opened 2026-04-04
feat(lib): add Unlimiformer-style kNN attention retrieval helpers

kayo09:
Implement the score-and-top-k pattern from Bertsch et al. (arXiv:2305.01625) for embedding vectors: dot-product attention scores with top-k key selection, plus multi-head and generic item ranking utilities. Add Bun tests and a test script for @repo/lib.


MaheshtheDev:
we don't need this in lib right now

PR #459, opened 2025-10-07
security: fix critical DOM-based XSS in browser extension

akhilesharora:
## Summary
Fixes critical DOM-based XSS vulnerability in browser extension content scripts for ChatGPT and Claude integrations.

## Problem
The extension was using `innerHTML` to inject API responses into the DOM without sanitization. If the Supermemory API is compromised or a MITM attack occurs, malicious JavaScript could execute in the context of ChatGPT/Claude, leading to:
- Session hijacking
- Cookie/token theft  
- Conversation exfiltration
- Complete account takeover

**Vulnerable code:**
```typescript
promptTextarea.innerHTML = `${promptTextarea.innerHTML} ${storedMemories}`
```

## Solution
Replace `innerHTML` with `textContent` which automatically escapes HTML:
```typescript
const memoriesContainer = document.createElement('div')
memoriesContainer.textContent = storedMemories
promptTextarea.appendChild(memoriesContainer)
```

## Files Changed
- `apps/browser-extension/entrypoints/content/chatgpt.ts` (line 656)
- `apps/browser-extension/entrypoints/content/claude.ts` (line 514)

## Testing
- [x] Tested with malicious payloads - XSS prevented
- [x] Verified functionality unchanged
- [x] HTML properly escaped in output

## Impact
**Severity:** Critical  
**Affected:** All browser extension users  
**Fix:** 3 lines per file

This is a security-critical fix that should be merged and released as soon as possible.

PR #838, opened 2026-04-08
security(browser-extension): fix DOM-based XSS vulnerability

akhilesharora:
## Summary
Fixes critical DOM-based XSS vulnerability in browser extension content scripts.

Continuation of #459 with feedback addressed.

## Changes
- Remove HTML wrappers (`<div>`, `<br>`) from stored memories - fixes the visible HTML tags issue
- Replace `innerHTML` with `document.createTextNode()` - prevents XSS
- Fix all three content scripts: chatgpt.ts, claude.ts, t3.ts

## Feedback Addressed
- ✅ Fixed UI issue with `<div>` tags appearing in prompt (MaheshtheDev)
- ✅ Fixed t3.ts which was missed in original PR
- ✅ Using `createTextNode()` instead of wrapper elements

### DIFF
--- a/apps/browser-extension/entrypoints/content/chatgpt.ts
+++ b/apps/browser-extension/entrypoints/content/chatgpt.ts
@@ -159,7 +159,7 @@ async function getRelatedMemoriesForChatGPT(actionSource: string) {
 		if (response?.success && response?.data) {
 			const promptElement = document.getElementById("prompt-textarea")
 			if (promptElement) {
-				promptElement.dataset.supermemories = `<div>Supermemories of user (only for the reference): ${response.data}</div>`
+				promptElement.dataset.supermemories = `\n\nSupermemories of user (only for the reference): ${response.data}`
 				console.log(
 					"Prompt element dataset:",
 					promptElement.dataset.supermemories,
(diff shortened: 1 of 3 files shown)


MaheshtheDev:
@akhilesharora can you share demo with this new changes?

MaheshtheDev:
@akhilesharora any update?

akhilesharora:
> @akhilesharora any update?

Tested this by saving memories with HTML payloads (``) through both the UI and direct API calls. Currently the XSS doesn't actually fire - the backend's AI processing rewrites everything into clean text before it reaches the extension.

For example, saving ` My favorite color is blue` via `/v3/documents` comes back from `/v4/search` as just `"John Doe's favorite color is blue"` - the HTML is gone.

**To reproduce:**
```bash
# Save a memory with HTML payload
curl -X POST 'https://api.supermemory.ai/v3/documents' \
  -H 'Authorization: Bearer YOUR_API_KEY' \
  -H 'Content-Type: application/json' \
  -d '{"content": "<img src=x onerror=alert(document.cookie)> My favorite color is blue"}'

# Wait ~20 seconds for processing, then search
curl -X POST 'https://api.supermemory.ai/v4/search' \
  -H 'Authorization: Bearer YOUR_API_KEY' \
  -H 'Content-Type: application/json' \
  -d '{"q": "favorite color"}'

# Returns clean text, HTML stripped by AI processing
```

**That said, that's just a side effect of the AI summarization, not intentional sanitization.** The extension code itself doesn't validate anything and passes whatever comes back from the API straight to innerHTML. If the backend processing ever changes, this becomes exploitable.

The createTextNode fix is the right defensive approach regardless of what the backend does - just wanted to share what came up while integrating the SDK.

ishaanxgupta (approved):

PR #842, opened 2026-04-09
perf: scope message equality check to content+parts instead of full object

ved015:
## Summary
- Replace `JSON.stringify(entireMessage)` with targeted field comparisons in `areUIMessageArraysEqual`
- First compare `content` string directly (O(n) on string length, no allocation)
- Fall back to `JSON.stringify(parts)` only  skips metadata fields (`createdAt`, `annotations`, `experimental_attachments`) that never change during streaming

## Why
`areUIMessageArraysEqual` is called on every message state update during chat streaming. Serializing the entire `UIMessage` object (including all metadata) on every comparison is expensive especially for messages with large tool invocations.

@MaheshtheDev please have a look

### DIFF
--- a/apps/web/stores/chat.ts
+++ b/apps/web/stores/chat.ts
@@ -27,9 +27,9 @@ export function areUIMessageArraysEqual(
 			return false
 		}
 
-		// Compare the entire message using JSON serialization as a fallback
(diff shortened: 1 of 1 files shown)


MaheshtheDev (approved):
LGTM

PR #846, opened 2026-04-10
perf(ui): convert message pairing to O(n) forward pass in chain-of-thought

ishaanxgupta:
## Summary
Optimizes the message pairing algorithm in the `ChainOfThought` component.

Previously, for every user message in the chat thread, the component created a new sliced array and iterated forward with `.find()` to locate the next assistant message. This resulted in an $O(n^2)$ time complexity and repeated garbage collection, which becomes noticeably sluggish on long conversations.

## Changes
- Removed the `.slice(i + 1).find(...)` nested iteration inside the `for` loop.
- Introduced a `lastUserPair` state tracker to map the first subsequent assistant message directly to the most recent user message in a single $O(n)$ forward pass.


MaheshtheDev (approved):
thanks for contritbution

PR #847, opened 2026-04-10
perf: wrap list/grid components in React.memo to skip unnecessary re-renders

ved015:
## Summary
- Wrap `UserMessage`, `WebsitePreview`, `YoutubePreview`, and `FilePreview` in `React.memo()` to prevent unnecessary re renders
- No behavioral or visual changes

## Why
These components render inside lists/grids that can have 50+ items. Any parent state change (eg new streaming token in chat, selection toggle in grid) causes all of them to re render even though their props have not changed.

## Impact
- **Skips re-renders** for unchanged user messages during assistant streaming
- **Skips re-renders** for unchanged preview cards in the document masonry grid
- Particularly noticeable in long conversations and large document grids

### DIFF
--- a/apps/web/components/chat/message/user-message.tsx
+++ b/apps/web/components/chat/message/user-message.tsx
@@ -1,5 +1,6 @@
 "use client"
 
+import { memo } from "react"
 import { Copy, Check } from "lucide-react"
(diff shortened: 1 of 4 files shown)


MaheshtheDev (approved):
looks great to me

PR #851, opened 2026-04-15
docs: clarify region requirement for S3-compatible endpoints

davidmyriel:
Hey there! Small docs fix for the S3 connector.

The `region` field is marked required in the config table, but the S3-compatible examples (MinIO, R2, Spaces) never say what to put there. Since those services don't use AWS regions, users either guess an AWS region or get stuck wondering why the field is asking for something that doesn't apply to their setup.

This PR:

- Adds a one-liner explaining that `region: 'auto'` works for S3-compatible endpoints (the value is still needed for request signing, but the service ignores it)
- Updates the MinIO example to use `auto` instead of `us-east-1`
- Annotates the R2 and Spaces endpoint lines with the same hint
- Adds Tigris to the list of S3-compatible providers while I was in there — same endpoint pattern as the others

No code changes, no new pages, just making an existing doc a bit clearer. Happy to adjust wording if you'd rather keep the scope tighter.

Thanks for taking a look!

### DIFF
--- a/apps/docs/connectors/s3.mdx
+++ b/apps/docs/connectors/s3.mdx
@@ -4,7 +4,7 @@ description: "Connect Amazon S3 or S3-compatible storage to sync files into your
 icon: "aws"
 ---
 
-Connect Amazon S3 buckets or S3-compatible storage services (MinIO, DigitalOcean Spaces, Cloudflare R2) to sync files into your Supermemory knowledge base.
+Connect Amazon S3 buckets or S3-compatible storage services (MinIO, DigitalOcean Spaces, Cloudflare R2, Tigris) to sync files into your Supermemory knowledge base.
 
 <Note>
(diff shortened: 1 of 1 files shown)


ved015:
Hi @davidmyriel, the PR looks good overall 

I had one small suggestion around the wording for the `region` explanation. Instead of mentioning “or any non-empty value”, maybe we can make it a bit more precise with something like:

* Use `region: 'auto'` for S3-compatible providers that do not expose AWS-style regions.
* For providers like MinIO or R2, use `region: 'auto'`.

Rest of the changes look great

ved015:
@davidmyriel can you pls resolve the conflicts so that we can close this

davidmyriel:
@ved015 Vedant, I'm very sorry for the delay. I just sent over the changes as requested. Thank you!

ved015 (approved):

ved015:
@davidmyriel thank you for your contribution

PR #867, opened 2026-04-20
fix: derive empty-state chat suggestions from user's memories

Sunny-ybe:
What

Replace hardcoded empty-state suggestions in the chat sidebar with prompts derived from the user's actual memories.

 Why

The old `DEFAULT_SUGGESTIONS` referenced content-specific strings like "My Gita" — a new user immediately sees prompts about content they don't own, which breaks trust before the first interaction. The `emptyStateSuggestions` prop on `ChatSidebar` already existed but was never sourced; this PR completes that.

 How

- extractMemoryLabel + deriveSuggestionsFromMemories — pure helper functions, easily unit-testable
- useDynamicSuggestions hook — fetches up to 20 memories via `POST /v3/documents/documents` (scoped by `containerTags`) and returns suggestions with this fallback chain:
- 
- Parent-supplied emptyStateSuggestions prop always wins and existing callers unaffected
- Skeleton loader at identical dimensions to prevent layout shift when suggestions swap in

Notes

- No new endpoints. Reuses the same route the memory grid calls.
- No LLM call in the generation path. Template-based; LLM-generated is a natural follow-up.
- Hook + helpers are decoupled from the sidebar UI, so they survive the upcoming chat refactor.


MaheshtheDev:
@Sunny-ybe can you resolve the conflict?

PR #913, opened 2026-05-09
fix: Implement the onboarding account lookup

ishaanxgupta:
Added an onboarding account verification flow for X and LinkedIn profile inputs. The onboarding UI now debounces profile entry, shows an inline loading state, and displays verified, not found, or unable-to-verify status before allowing the user to continue. A new `GET /api/onboarding/account-status` route normalizes X/Twitter handles and LinkedIn profile URLs, verifies X accounts through Twitter’s oEmbed endpoint to avoid false positives from X’s generic profile shell, and handles LinkedIn cautiously by returning an unverified state when LinkedIn blocks server-side checks instead of incorrectly marking the profile as found. Confirmed missing accounts block submission, while uncertain verification errors still allow the user to continue.

image

image

### DIFF
--- a/apps/web/app/(app)/onboarding/page.tsx
+++ b/apps/web/app/(app)/onboarding/page.tsx
@@ -467,6 +467,7 @@ export default function OnboardingPage() {
 					found?: boolean
 					handle?: string
 					reason?: string
+					verified?: boolean
 				} = await response.json().catch(() => ({}))
 
 				if (controller.signal.aborted) return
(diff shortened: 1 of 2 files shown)


ved015 (approved):

apps/web/app/(app)/onboarding/page.tsx
```diff
@@ -416,6 +435,97 @@ export default function OnboardingPage() {
 		return () => clearTimeout(t)
 	}, [])
 
+	useEffect(() => {
```
MaheshtheDev, line 420: can we not make it as hook?

ishaanxgupta: yes done!

MaheshtheDev (approved):

PR #919, opened 2026-05-10
fix: fetch chat history when opening recent conversations sheet

ved015:
## Problem
Clicking "Recent Conversations" showed empty list because:
1. `fetchThreads()` was only called in Sheet's `onOpenChange`, which doesn't trigger when opened via toolbar button
2. Auto-restore effect tried to load threads from URL even for newly created chats (causing 404s)

## Solution
- Add `useEffect` that triggers `fetchThreads()` whenever `isHistoryOpen` becomes true
- Fix auto-restore to only load threads that existed in URL on initial mount

Screenshot 2026-05-10 at 12 17 15 PM

### DIFF
--- a/apps/web/components/chat/index.tsx
+++ b/apps/web/components/chat/index.tsx
@@ -399,6 +399,12 @@ export function ChatSidebar({
 		}
 	}, [selectedProject])
 
+	useEffect(() => {
+		if (!isHistoryOpen) return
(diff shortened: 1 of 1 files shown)


ishaanxgupta (approved):

MaheshtheDev (approved):
LGTM

PR #924, opened 2026-05-10
fix: improve settings sidebar overflow

ishaanxgupta:
## Summary
- Make the desktop settings sidebar independently scrollable with a hidden scrollbar
- Tighten desktop sidebar spacing so all navigation and danger actions fit better on laptop screens
- Use `headerDisplayName` for the settings orb label so magic-link users show `Name's supermemory`. Extracted the name from the email only because user's object is empty for magic link users.

## before:
image

## after:
image


ved015 (approved):
Nice catch lgtm

PR #964, opened 2026-05-18
Improve quick note editor interactions & fix filter chips responsiveness

ishaanxgupta:
On phone, the filter chips now get their own horizontally scrollable row, so All, Notes, Tweets, Web Pages, and any future connection chips can fit without squeezing. The Grid / Timeline segmented control stays grouped on the next row and aligned to the right.

The QuickNoteCard now uses the same TipTap TextEditor as the full-screen note, so typing / should show the slash-command options like headings. It also expands from 120px to 188px when focused or when it has draft content, then collapses again on blur if empty.

I also updated TextEditor so it can be used inline safely: it now supports editable={false} while saving, and it syncs external content resets without constantly resetting the editor while typing.

### DIFF
--- a/apps/web/components/add-document/note.tsx
+++ b/apps/web/components/add-document/note.tsx
@@ -20,11 +20,9 @@ export function NoteContent({
 	const [content, setContent] = useState(initialContent ?? "")
 	const [seededContent] = useState(initialContent || undefined)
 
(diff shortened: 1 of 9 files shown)


ved015:
@ishaanxgupta pls have a look at the suggestions

ved015 (approved):

PR #971, opened 2026-05-19
feat(extension): Add support across gemini & claude, live tool tip, auto complete support

ishaanxgupta:
extension works across chatgpt, gemini, claude. there is a live marker now with tool tip that shows supermemory is connected, the welcome msg is revamped fully. 
The autocomplete works as expected.

here is a photo of the welcome page:
image


Dhravya (changes requested):
Two things before this can merge:

1. `CHATGPT_DEBUG = true` ships verbose logging to every user — flip it to `false` (the `debugChatGPT` helper already no-ops when it's off), or derive it from an env/build flag.
2. There are a few raw `console.log`s outside the debug helper (e.g. `console.log("Prompt element dataset:", memoryText)`) that log user prompt content — route them through `debugChatGPT` or drop them; prompt text shouldn't hit the console in production.

Also please rebase on main — the branch is from May and mergeability is currently unknown.

ishaanxgupta:
@Dhravya can you review now?

PR #973, opened 2026-05-19
Improve memory graph

Dhravya:
## Summary
- add a container tag selector to the memory graph playground, including a proxy for `/v3/container-tags/list`
- improve dense graph readability with stable append placement, wider memory clustering, dynamic zoom bounds, and relation-edge level-of-detail
- add cluster-colored nodes, clearer update-chain markers, stronger focused update edges, and an expanded legend explaining colors/edge semantics
- make the playground converter match current API response shapes (`memoryEntries`, `type`, relation fields)

## Verification
- `bun --filter @supermemory/memory-graph test`
- `bun run build` in `packages/memory-graph`
- `bun run build` in `apps/memory-graph-playground`
- `git diff --cached --check`


ishaanxgupta:
the graph looks like this in full view:
image
image

ishaanxgupta:
Added low-zoom rendering optimization:

- samples dense document-memory edges
- fades/simplifies low-zoom edges
- draws huge memory sets as batched colored points instead of expensive detailed nodes

PR #993, opened 2026-05-23
fix(openai-middleware): prevent double-wrapping and cross-tenant memory leakage on shared client

devteamaegis:
## Problem

`createOpenAIMiddleware` / `withSupermemory` mutates the caller's `OpenAI` client in-place by overwriting `openaiClient.chat.completions.create`. In the most common server-side usage pattern — one shared `OpenAI` instance, `withSupermemory` called per-request with the current user's `containerTag` — this creates two bugs:

### 1. Double memory injection (latency / cost)

Every call after the first captures the **already-wrapped** `create` as `originalCreate`, stacking wrappers:

```
request 1  → user-A wrapper installed
request 2  → user-B wrapper installed ON TOP of user-A wrapper
             originalCreate (for user-B) = user-A's createWithMemory

completion → user-B injects memories → calls user-A wrapper → injects again → real API call
```

Each completion now hits the Supermemory profile API twice, doubling latency and token cost.

### 2. Cross-tenant memory leakage (security)

The inner wrapper (user-A's) runs with `containerTag = "user-A"`, so user-B's completion receives user-A's memories injected into its system prompt.

```typescript
// Server handler — shared client
const openai = new OpenAI({ apiKey: process.env.OPENAI_API_KEY })

app.post("/chat", async (req, res) => {
  const client = withSupermemory(openai, {  // ← BUG: mutates shared `openai`
    containerTag: req.user.id,
    customId: req.body.conversationId,
  })
  // After the second request, `openai.chat.completions.create` is a
  // double-wrapped function. Every completion includes memories from
  // both the current user AND the previous request's user.
  const response = await client.chat.completions.create(...)
})
```

## Fix

Attach the original SDK method to every installed wrapper via a private `Symbol` (`ORIGINAL_CREATE_SYM`). Before capturing `originalCreate`, unwrap through the symbol if present. This ensures no matter how many times `createOpenAIMiddleware` is called on the same client, the innermost call is always the real SDK `create`:

```typescript
const ORIGINAL_CREATE_SYM = Symbol("supermemory.originalCreate")

// Capture (unwrap if previously wrapped):
const currentCreate = openaiClient.chat.completions.create
const originalCreate = currentCreate[ORIGINAL_CREATE_SYM] ?? currentCreate

// After building createWithMemory, stamp the original on it:
createWithMemory[ORIGINAL_CREATE_SYM] = originalCreate

// Then install as before:
openaiClient.chat.completions.create = createWithMemory
```

The same pattern is applied to the Responses API path.

## Tests

Three unit tests added (`middleware.test.ts`, no network calls — `fetch` is stubbed):

| Test | Asserts |
|---|---|
| Two calls → real `create` invoked once | No double-wrapping |
| Three calls → real `create` invoked once | Idempotent for N calls |
| Latest call wins | `containerTag` from second call used, not first |

```
✓ calling createOpenAIMiddleware twice ... invokes the real SDK create once
✓ calling createOpenAIMiddleware three times ... invokes the real SDK create once  
✓ the wrapper installed by the second call replaces the first (latest wins)
```

## Notes

- The public API surface is unchanged — `withSupermemory` still returns the same mutated client.
- A longer-term fix would be to return a `Proxy` instead of mutating the input, but that's a breaking change. This Symbol guard is backward-compatible.

### DIFF
--- a/packages/tools/src/openai/middleware.test.ts
+++ b/packages/tools/src/openai/middleware.test.ts
@@ -1,0 +1,180 @@
+/**
+ * Unit tests for createOpenAIMiddleware — no network calls required.
+ *
+ * These tests verify the double-wrapping protection added to guard against
+ * cross-tenant memory leakage when the same OpenAI client is passed to
+ * `createOpenAIMiddleware` (or `withSupermemory`) multiple times, which is
+ * the common pattern in multi-user server applications.
+ */
+
+import { describe, it, expect, vi, beforeEach } from "vitest"
+import type OpenAI from "openai"
+
+// The Supermemory SDK client constructor throws when SUPERMEMORY_API_KEY is
+// absent. Provide a fake key here — the actual value is irrelevant because
+// all network calls are intercepted by a `vi.stubGlobal("fetch", ...)` mock.
+process.env.SUPERMEMORY_API_KEY = "sm_test_fake_key_for_unit_tests"
+
+// We import the internal implementation directly to avoid the SUPERMEMORY_API_KEY
+// check in the public `withSupermemory` wrapper.
+import { createOpenAIMiddleware } from "./middleware"
+
+// ---------------------------------------------------------------------------
+// Minimal OpenAI client stub
+// ---------------------------------------------------------------------------
+
+/** Number of times the real (innermost) `create` was called. */
+let realCreateCallCount = 0
+
+/**
+ * Build a minimal mock OpenAI client whose `chat.completions.create` is a
+ * spy tracking how many times the underlying SDK method is invoked.
+ */
+function makeOpenAIStub(): OpenAI {
+	realCreateCallCount = 0
+
+	const realCreate = vi.fn(async () => {
+		realCreateCallCount++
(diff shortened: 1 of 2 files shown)


ishaanxgupta (changes requested):
hey @devteamaegis thank you for the contribution.
This fixes stacked wrappers, but I don’t think it fully fixes the cross-tenant shared-client case described in the PR.
Because `withSupermemory` still mutates and returns the same `OpenAI` object, any previously returned reference observes the latest installed wrapper:

```ts
const clientA = withSupermemory(sharedOpenAI, { containerTag: "user-A", customId: "a" })
const clientB = withSupermemory(sharedOpenAI, { containerTag: "user-B", customId: "b" })

await clientA.chat.completions.create(...) // uses user-B wrapper, not user-A
```
So the Symbol guard prevents wrapper stacking, but it makes latest call wins globally on the shared instance. That is still unsafe for overlapping requests or any async flow that keeps a wrapped client reference.

ishaanxgupta:
Closing for now!

PR #994, opened 2026-05-24
fix(onboarding): parse X URLs with www. and bare-domain prefixes

vimzh:
The extractHandle helper in the onboarding research route relies on a few naive .replace() calls to strip protocol+domain prefixes, then splits on `/` and `?`. Anything other than the exact form `https://x.com/handle` or `https://twitter.com/handle` produces garbage. Confirmed locally:

- `https://www.x.com/vansh` returns `"https:"`
- `x.com/vansh` returns `"x.com"`
- `https://mobile.twitter.com/foo` returns `"https:"`

That junk handle then flows into the Grok call, so we burn an LLM request and the user gets back research about the wrong "handle".

This rewrites extractHandle using the URL constructor, mirroring the parseXAccount pattern that already exists in `apps/web/app/api/onboarding/account-status/route.ts`. A regex fallback handles inputs the URL constructor refuses. After extraction the handle is validated against the 1-15 char alphanumeric+underscore X handle rule, so anything still malformed returns a clean 400 instead of a wasted Grok call.

The parsing logic is now duplicated between this route and account-status. A shared helper would be nice but felt out of scope for this one.

### DIFF
--- a/apps/web/app/api/onboarding/research/route.ts
+++ b/apps/web/app/api/onboarding/research/route.ts
@@ -7,16 +7,28 @@ interface ResearchRequest {
 	email?: string
 }
 
-function extractHandle(url: string): string {
-	const cleaned = url
-		.toLowerCase()
-		.replace("https://x.com/", "")
-		.replace("https://twitter.com/", "")
-		.replace("http://x.com/", "")
-		.replace("http://twitter.com/", "")
-		.replace("@", "")
+function extractHandle(input: string): string {
+	const trimmed = input.trim()
+	if (!trimmed) return ""
 
-	return (cleaned.split("/")[0] ?? cleaned).split("?")[0] ?? cleaned
+	let handle = trimmed.replace(/^@+/, "")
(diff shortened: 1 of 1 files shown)


ishaanxgupta (changes requested):
`isXHost()` allows every `*.x.com` / `*.twitter.com` host, and `extractHandle()` then treats the first path segment as the handle. For example:

```ts
https://developer.x.com/en/docs -> "en"
https://help.x.com/en/using-x -> "en"
```
Can we restrict the accepted hosts to actual profile hosts, e.g. x.com, twitter.com, www.x.com, www.twitter.com, and maybe mobile.twitter.com, and make the fallback regex follow the same host rules?

MaheshtheDev (approved):

MaheshtheDev:
@vimzh did you get a chance look at it?

ishaanxgupta (approved):

ishaanxgupta:
Hi @vimzh thanks for the contribution

PR #995, opened 2026-05-24
fix(onboarding): fail fast when EXA_API_KEY is missing

vimzh:
The extract-content route currently uses `process.env.EXA_API_KEY ?? ""` as the auth header. Deploying without EXA_API_KEY set sends a request to Exa with an empty key, Exa returns an auth failure, and the client gets back a generic 500 "Failed to fetch content from Exa API". From the outside that looks identical to any upstream error, which makes self-hosted setups frustrating to debug.

This reads EXA_API_KEY once at the top of the handler. If it's missing we log a specific message and return a 500 with "Content extraction service is not configured" so the cause is obviously a config issue rather than an upstream hiccup. The downstream fetch then uses the validated key with no fallback.

Kept the status as 500 to stay consistent with the file's existing 500 for upstream errors. 503 could be argued for a misconfig but introducing a new status code felt like scope creep here.

### DIFF
--- a/apps/web/app/api/onboarding/extract-content/route.ts
+++ b/apps/web/app/api/onboarding/extract-content/route.ts
@@ -9,14 +9,19 @@ interface ExaApiResponse {
 	results: ExaContentResult[]
 }
 
+const exaApiKey = process.env.EXA_API_KEY
+if (!exaApiKey) {
+	console.error(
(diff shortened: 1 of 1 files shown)


ishaanxgupta:
Hey @vimzh thank you for the contribution. LGTM

ishaanxgupta (approved):

PR #996, opened 2026-05-24
fix(tweet-preview): guard JSON.parse to avoid crashing the modal

vimzh:
TweetPreview runs `JSON.parse(data)` on the data prop with no try/catch. If a document's stored Twitter metadata is ever malformed (truncated JSON, corrupted serialization, exotic stringified values like `"null"` or `"123"`), the parse throws and the React tree rendering the document modal or memory card crashes.

This wraps the parse in a small `parseTweetData` helper that returns null when:

- The input is empty or null
- JSON.parse throws
- The parsed value isn't a plain object (covers arrays, which would otherwise slip past a `typeof === "object"` check and crash deeper in enrichTweet)

When parsing returns null the component renders nothing instead of crashing. Also widened the prop type from `Tweet` to `Tweet | string` to match what both call sites already pass via `as Tweet` / `as unknown as Tweet`.

One thing worth flagging for reviewers: in `apps/web/components/document-modal/content/tweet.tsx` the TweetPreview branch is entered whenever tweetMetadata is truthy, so malformed metadata now renders an empty container rather than the "Tweet preview unavailable" fallback further down that file. That is strictly better than crashing but not the friendliest UX. Happy to move the validation up into TweetContent in a follow-up if you would prefer the explicit fallback.

### DIFF
--- a/apps/web/components/document-cards/tweet-preview.tsx
+++ b/apps/web/components/document-cards/tweet-preview.tsx
@@ -1,6 +1,6 @@
 "use client"
 
-import { Suspense } from "react"
+import { Suspense, useMemo } from "react"
 import type { Tweet } from "react-tweet/api"
 import { TweetBody, enrichTweet, TweetSkeleton } from "react-tweet"
 import { cn } from "@lib/utils"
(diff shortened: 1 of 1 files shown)


ishaanxgupta (changes requested):
`parseTweetData()` only checks that the parsed value is a non-array object with a `"user"` key:

```ts
"user" in value
```

But `enrichTweet(parsedTweet)` still assumes the full `Tweet` shape exists. So valid JSON like these can still pass `isTweetLike()` and crash during render:

```ts
{"user": null}
{"user": {}}
{"user": {}, "text": null}
```

For example, `enrichTweet` reads fields like `tweet.user.screen_name`, `tweet.text`, `tweet.display_text_range`, and `tweet.entities.*`. If any of those are missing or malformed, we’re past the parse guard and the React tree can still throw.

ishaanxgupta:
This still lets invalid tweet data through. For example, `{"user":{"screen_name":"x"},"text":"hi","display_text_range":[0,2],"entities":{}}` passes `isTweetLike()`, but `enrichTweet()` expects `entities.hashtags`, `entities.user_mentions`, `entities.urls`, and `entities.symbols` to be arrays and will throw when iterating undefined. Can we validate/default those required arrays before returning `Tweet`?

ishaanxgupta:
LGTM!

ishaanxgupta (approved):

PR #999, opened 2026-05-24
fix(openai-sdk): don't inject empty system prompt when no memories are found

devteamaegis:
## Bug

In `packages/openai-sdk-python/src/supermemory_openai/middleware.py`, `add_system_prompt()` builds a `memories` string and then unconditionally modifies the message list — even when `memories` is empty (i.e. the profile and search results both returned nothing).

**What happens today when no memories are found:**

1. **Existing system prompt present** — the content becomes `"You are helpful. \n "` (trailing whitespace appended).
2. **No system prompt present** — a new `{"role": "system", "content": ""}` message is prepended to the list.

Both outcomes corrupt the conversation: the first wastes tokens with whitespace junk; the second injects a blank system turn that some models treat as an explicit instruction.

**Minimal reproduction:**
```python
# memories resolves to "" when profile/search return nothing
messages = [{"role": "user", "content": "Hello"}]
result = await add_system_prompt(messages, ...)
# Before fix: [{"role": "system", "content": ""}, {"role": "user", "content": "Hello"}]
# After fix:  [{"role": "user", "content": "Hello"}]   (unchanged)
```

## Fix

Add an early return before the modification block:

```python
if not memories:
    return messages
```

One line. No behaviour change when memories are present.

## Tests

Added `test_empty_memories_do_not_modify_messages` to `tests/test_middleware.py` covering both cases (messages with and without a pre-existing system prompt).

### DIFF
--- a/packages/openai-sdk-python/src/supermemory_openai/middleware.py
+++ b/packages/openai-sdk-python/src/supermemory_openai/middleware.py
@@ -193,6 +193,9 @@ async def add_system_prompt(
             },
         )
 
+    if not memories:
+        return messages
+
     if system_prompt_exists:
         logger.debug("Added memories to existing system prompt")
         return [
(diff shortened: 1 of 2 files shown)


ishaanxgupta:
Thanks for the contribution

ishaanxgupta (approved):

PR #1007, opened 2026-05-26
feat/granola connector

sreedharsreeram:


apps/web/components/granola-connect-modal.tsx
```diff
@@ -0,0 +1,222 @@
...
+		}
+	}, [open])
+
+	const connectMutation = useMutation({
+		mutationFn: async (key: string) => {
+			const response = await $fetch("@post/connections/:provider", {
+				params: { provider: "granola" },
+				body: {
+					containerTags,
+					metadata: { apiKey: key },
```
MaheshtheDev, line 57: can we encrpt while sending this and decrypt when we want to use?

sreedharsreeram: it does get encrypted/decrypted and the apikey gets saved in the accessToken column

MaheshtheDev (approved):

PR #1008, opened 2026-05-26
fix(pipecat): retain strong references to background storage tasks to prevent GC

devteamaegis:
## Problem

In `SupermemoryPipecatService.process_frame`, conversation messages are stored
via a fire-and-forget coroutine:

```python
asyncio.create_task(self._store_messages(unsent_messages))
```

The `Task` object returned by `asyncio.create_task` is immediately discarded.
Python's `asyncio` event loop only holds **weak references** to scheduled
tasks, so the garbage collector is free to destroy a task at any time — even
while it is still awaiting an I/O call inside `_store_messages`.  On a GC
cycle this can silently kill an in-flight `supermemory.memories.add()` request,
dropping messages with no error logged.

The `SupermemoryCartesiaAgent` in the same repo (cartesia-sdk-python) already
uses the correct pattern: it saves each task in a `_background_tasks` set and
removes it via `add_done_callback(discard)` once the coroutine finishes.

## Fix

Apply the same pattern to `SupermemoryPipecatService`:

1. Add `self._background_tasks: set = set()` in `__init__`.
2. Save every storage task in the set; remove via `add_done_callback`.
3. Clear the set in `reset_memory_tracking()`.

```python
# before
asyncio.create_task(self._store_messages(unsent_messages))

# after
task = asyncio.create_task(self._store_messages(unsent_messages))
self._background_tasks.add(task)
task.add_done_callback(self._background_tasks.discard)
```

## Tests

`tests/test_background_task_tracking.py` (all 5 pass):

- `test_background_tasks_set_exists` – attribute present after init
- `test_task_held_during_execution` – task is in the set while running, removed after
- `test_task_removed_after_completion` – set is empty once coroutine finishes
- `test_gc_cannot_collect_tracked_task` – forced `gc.collect()` mid-execution can't kill a tracked task
- `test_reset_clears_background_tasks` – `reset_memory_tracking()` empties the set

### DIFF
--- a/packages/pipecat-sdk-python/src/supermemory_pipecat/service.py
+++ b/packages/pipecat-sdk-python/src/supermemory_pipecat/service.py
@@ -118,6 +118,7 @@ class SupermemoryPipecatService(FrameProcessor):
         self._messages_sent_count: int = 0
         self._last_query: Optional[str] = None
         self._audio_frames_detected: bool = False
+        self._background_tasks: set = set()  # Prevent GC of fire-and-forget tasks
 
     async def _retrieve_memories(self, query: str) -> Dict[str, Any]:
         """Retrieve relevant memories from Supermemory.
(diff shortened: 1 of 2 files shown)


ishaanxgupta:
@devteamaegis did you get a chance to look on the review?

PR #1012, opened 2026-05-27
feat(zoom): zoom connector [needs testing]

sreedharsreeram:


Dhravya:
Thanks for the contribution! Holding this for now — the Zoom connector isn't released yet on the backend side, so we don't want to surface the option in the UI until it's live end-to-end. We'll revisit and get this merged once the Zoom integration ships.

One small ask in the meantime: the schema change makes `authLink`/`expiresIn` optional for *all* providers — once the Zoom endpoint returns an `authLink` like the other OAuth providers, that loosening shouldn't be needed and it'd be good to revert it so a missing link fails loudly instead of leaving the button stuck on "connecting".

ishaanxgupta:
Broken “Configure Zoom” screen
The UI accepts zoomClientId, zoomClientSecret, and zoomCustomKeyEnabled, but the API/database have none of these fields. The settings validator silently converts that payload to {}, while the UI displays “Settings saved.”
Either remove Zoom from the custom-credentials screen or implement the database fields, encryption, validation, provider selection, and per-app webhook secret handling.

Webhook events trigger too many connections
The query matches accountId OR hostEmail. Everyone in the same company Zoom account shares the account ID, so one recording can trigger a full sync for every connected employee. The payload’s exact host_id and meeting uuid are parsed but ignored. Zoom documents host_id as part of these recording events. Zoom webhook documentation
Store the Zoom user ID and target the exact connection and meeting.

OAuth can succeed without webhook identity
If /users/me fails, the exception is logged and connection creation continues without email or account ID. Initial sync may work, but future webhooks can never locate that connection. Zoom setup should fail if its identity lookup fails.

- getZoomUser() is unused and duplicates ZoomProvider.getUserInfo().
- The generic Provider.getUserInfo() type only declares email, forcing an unsafe cast for accountId. Extend the interface instead.
- Handle Zoom’s Retry-After response for 429 errors instead of using only a fixed 250 ms delay.
- Change the label to “Sync cloud-recorded Zoom meetings…” and mention that Zoom cloud recording/transcription is required; local/free-plan recordings are unsupported.

PR #1013, opened 2026-05-27
chore(web): bump @lobbyside/react to 0.5.0 and switch to org install

TropicolX:
Pins to the latest published version and replaces the per-widget id with the supermemory org id, so flipping which widget is live in the Lobbyside dashboard takes effect without a code change.

### DIFF
--- a/apps/web/components/next-app-research-cta.tsx
+++ b/apps/web/components/next-app-research-cta.tsx
@@ -13,7 +13,7 @@ const STORAGE_KEY = "sm_next_app_research_cta_dismissed_v1"
 
 const BOOK_CALL_HREF = "https://cal.com/supermemory/growth"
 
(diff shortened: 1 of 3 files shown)


ishaanxgupta (commented):
Could you regenerate `bun.lock` so this PR only contains the Lobbyside package change? The current lockfile diff pulls in unrelated workspace/dependency churn (`apps/raycast-extension`, Raycast/oclif/type-eslint entries, and shared package changes such as `ajv`/`ansi-escapes`), which makes this web-only PR change dependency resolution outside the intended scope. Also please run the repo formatter on `apps/web/package.json`; the new `@lobbyside/react` line has tab indentation while the surrounding JSON uses spaces.

MaheshtheDev:
invalid PR

PR #1022, opened 2026-05-29
add web upgrade plans UI

ishaanxgupta:
## Summary
- Add a blue Upgrade CTA in the web header immediately before Add memory
- Add a web-styled Free/Pro Plans section to the Billing settings tab using the same plan copy and Pro checkout flow as the existing billing UI.
- Add the Upgrade option to the mobile header menu so the action remains available on small screens.

image
image


MaheshtheDev (changes requested):
can we have better upgrade icon and button can be bit primary. now it is so blant like other buttons

MaheshtheDev (approved):
LGTM

PR #1028, opened 2026-05-31
fix(pipecat): guard against None profile in _retrieve_memories()

### LINKED ISSUE #1027: BUG: AttributeError when profile is None in _retrieve_memories()
## What breaks

`SupermemoryPipecatService._retrieve_memories()` in `packages/pipecat-sdk-python/src/supermemory_pipecat/service.py` accesses `response.profile.static` and `response.profile.dynamic` without first checking whether `response.profile` is `None`. The Supermemory API returns `None` for the `profile` field when the user has no stored memories (e.g. a brand-new user). This causes an unhandled `AttributeError` that crashes the memory retrieval path and surfaces to callers as a `MemoryRetrievalError` wrapping the underlying exception.

## Traceback

```
File "service.py", line 155, in _retrieve_memories
    "static": response.profile.static,
              ^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'static'
```

## How to trigger

Call `_retrieve_memories()` against a user account that has no stored memories. The API responds with `profile=None`, which the code does not guard against before attribute access.

devteamaegis:
## What's broken

`SupermemoryPipecatService._retrieve_memories()` crashes with `AttributeError: 'NoneType' object has no attribute 'static'` when the Supermemory API returns `None` for the `profile` field. This happens for brand-new users who have no stored memories. The crash propagates up as a `MemoryRetrievalError` wrapping the `AttributeError`, breaking the entire memory retrieval path on first use.

## Why it happens

Lines 155–156 of `service.py` access `response.profile.static` and `response.profile.dynamic` unconditionally, but the API contract allows `profile` to be `None`.

## Fix

Added a `None` check before accessing each attribute, defaulting to an empty list when `response.profile is None`. The change is two lines in `_retrieve_memories()`.

## Test

Added `tests/test_service_none_profile.py` which asserts that the fixed guard returns empty lists when `profile` is `None`, and that it still returns the correct values when `profile` is populated.

Fixes #1027

### DIFF
--- a/packages/pipecat-sdk-python/src/supermemory_pipecat/service.py
+++ b/packages/pipecat-sdk-python/src/supermemory_pipecat/service.py
@@ -150,10 +150,13 @@ class SupermemoryPipecatService(FrameProcessor):
             if response.search_results and response.search_results.results:
                 search_results = response.search_results.results
 
+            profile_static = response.profile.static if response.profile is not None else []
+            profile_dynamic = response.profile.dynamic if response.profile is not None else []
+
             return {
                 "profile": {
-                    "static": response.profile.static,
-                    "dynamic": response.profile.dynamic,
+                    "static": profile_static,
+                    "dynamic": profile_dynamic,
                 },
                 "search_results": search_results,
             }
--- a/packages/pipecat-sdk-python/tests/test_service_none_profile.py
+++ b/packages/pipecat-sdk-python/tests/test_service_none_profile.py
@@ -1,0 +1,32 @@
+"""Test that _retrieve_memories handles a None profile without raising AttributeError."""
+
+import pytest
+
+
+def test_retrieve_memories_none_profile_returns_empty_lists():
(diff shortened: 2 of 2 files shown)


ishaanxgupta (changes requested):
Can we make the test call `_retrieve_memories()` with a mocked `_supermemory_client.profile()` response where `profile = None`? Right now the test duplicates the guard logic instead of exercising the service method, so it does not protect the actual regression.

ishaanxgupta (commented):
The regression test does not exercise `SupermemoryPipecatService._retrieve_memories()`; it defines fake classes and reimplements the same ternary expressions in the test body. That means the test would still pass if `service.py` regressed back to `response.profile.static`. Please update the test to instantiate/load the service with a mocked `_supermemory_client.profile()` returning `profile=None` and assert the actual `_retrieve_memories()` result.

ishaanxgupta:
Closing for now

PR #1031, opened 2026-06-01
feat(mcp): add privacy-safe recall receipts

### LINKED ISSUE #985: Expose privacy-safe memory retrieval receipts
Supermemory is positioning the right problem very clearly: memory is fragmented across Cursor, Claude, VS Code, Claude Code, Cline, etc., and MCP can make recall portable.

One observability/debugging gap I keep seeing in shared-memory workflows is: after a cross-client recall happens, what privacy-safe evidence can a user or maintainer inspect/share without dumping the raw memory, prompt, transcript, or private repo text?

For example, when someone says “Claude did not remember the PDF I added in Cursor” or “the wrong project memory was retrieved”, the useful artifact is not the memory body. It is a small receipt of what the memory layer returned and whether the client/harness loaded it.

A minimal receipt shape could be:

```text
memory.search.returned
  memory.provider = supermemory
  client.name = cursor | claude-code | vscode | ...
  project.id_hash
  query.hash                 # not raw query by default
  result.count
  result.ids_hash
  result.score_bucket[]
  result.content_hash[]      # hash of returned chunks, no raw memory text
  snapshot.id/hash
  latency_ms

context.input.loaded         # emitted by client/harness if it entered the session
  source.uri = supermemory://...
  delivered.hash
  estimated_chars
  activation = mcp_search | automatic_recall | manual
  duplicate.suppression_policy
```

That separation matters because the MCP server can prove “these memories were returned”, while the client/harness proves “this context actually entered the model/session”. Both can stay privacy-first.

Why I think this is useful:

- user bug reports can include receipts without exposing private memories;
- maintainers can distinguish “search returned nothing” vs “client did not load returned memory”;
- teams can audit cross-client memory usage without enabling raw prompt/memory logging;
- it creates a bridge to OpenTelemetry-style agent traces without making Supermemory responsible for the entire agent runtime.

I have a small public fixture exploring this general receipt pattern here, if useful as prior art: https://github.com/caioribeiroclw-pixel/pluribus/tree/main/examples/context-input-evidence

The concrete ask is small: would Supermemory consider an optional debug/export mode that emits privacy-preserving memory search/result receipts, or is this better as an external wrapper around the MCP client/server boundary?

cat0825:
## Summary

Adds an optional `includeReceipt` flag to the MCP `recall` tool. When enabled, recall responses include a privacy-safe retrieval receipt in `structuredContent.receipt`.

The receipt helps users and maintainers debug cross-client memory retrieval without exposing raw memory text, raw queries, project names, transcripts, or private repository content.

Closes #985.

## What changed

- Added `includeReceipt` to the MCP recall schema.
- Added `createRetrievalReceipt()` helper.
- Receipt includes:
  - hashed query
  - hashed project/container tag
  - hashed result IDs
  - hashed returned memory/chunk content
  - result count / total
  - score buckets
  - latency
  - MCP client name/version when available
  - profile static/dynamic counts when profile recall is used
- Added tests verifying raw private values are not exposed.

## Validation

- `bunx biome check --write apps/mcp/src/server.ts apps/mcp/src/retrieval-receipt.ts apps/mcp/src/retrieval-receipt.test.ts`
- `bunx vitest run src/retrieval-receipt.test.ts`
- `bun run --cwd apps/mcp build:ui`

`bun run check-types` currently fails on existing unrelated `@supermemory/tools` type errors.

### DIFF
--- a/apps/mcp/src/retrieval-receipt.test.ts
+++ b/apps/mcp/src/retrieval-receipt.test.ts
@@ -1,0 +1,58 @@
+import { describe, expect, it } from "vitest"
+import { createRetrievalReceipt, toScoreBucket } from "./retrieval-receipt"
+
+describe("retrieval receipts", () => {
+	it("hashes private values instead of exposing raw query, project, ids, or content", async () => {
+		const receipt = await createRetrievalReceipt({
+			query: "private cardiology PDF notes",
+			containerTag: "secret-client-project",
+			clientInfo: { name: "claude-code", version: "1.2.3" },
+			results: [
+				{
+					id: "mem_sensitive_1",
+					similarity: 0.87,
+					text: "Patient-specific private content",
+				},
+			],
+			total: 3,
+			latencyMs: 42,
+			profile: { staticCount: 2, dynamicCount: 1 },
+		})
+
+		expect(receipt).toMatchObject({
+			event: "memory.search.returned",
+			provider: "supermemory",
+			source: "mcp",
+			activation: "mcp_recall",
+			client: { name: "claude-code", version: "1.2.3" },
+			result: {
+				count: 1,
+				total: 3,
+				scoreBuckets: ["0.8-0.9"],
+			},
+			profile: { staticCount: 2, dynamicCount: 1 },
+			latencyMs: 42,
+			hashAlgorithm: "sha256-prefix-16",
+		})
+
+		expect(receipt.queryHash).toHaveLength(16)
+		expect(receipt.projectIdHash).toHaveLength(16)
+		expect(receipt.result.idsHash).toEqual([
+			expect.stringMatching(/^[a-f0-9]{16}$/),
+		])
+		expect(receipt.result.contentHashes).toEqual([
+			expect.stringMatching(/^[a-f0-9]{16}$/),
+		])
+		expect(JSON.stringify(receipt)).not.toContain("private cardiology")
+		expect(JSON.stringify(receipt)).not.toContain("secret-client-project")
+		expect(JSON.stringify(receipt)).not.toContain("mem_sensitive_1")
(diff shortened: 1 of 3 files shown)


ishaanxgupta (changes requested):
I don't think plain deterministic `sha256` prefixes are enough to call this privacy-safe. Query text, project/container tags, memory IDs, and especially short memory content can be dictionary-guessed offline, and the same private value is linkable across receipts forever.

cat0825:
Good catch, you're right that a plain deterministic SHA-256 prefix isn't privacy-safe for these inputs. I've reworked the hashing in `createRetrievalReceipt`:

- Each receipt now generates a fresh 32-byte random salt and keys an **HMAC-SHA256** with it. The salt is never emitted in the receipt.
- Without the salt, the low-entropy values (query, container/project tag, memory IDs, short content) can no longer be dictionary-guessed or precomputed offline.
- Because the salt is regenerated per receipt, the same private value produces a different token every time, so values are no longer linkable across receipts.
- Equality is still preserved *within* a single receipt (e.g. duplicate IDs/content in one result set), which is what the debugging use case needs.

`hashAlgorithm` is now reported as `hmac-sha256-ephemeral-salt-prefix-16`, and I added a regression test covering both properties (intra-receipt equality and cross-receipt non-linkability). Let me know if you'd prefer a different trade-off here, e.g. dropping content hashes entirely or making the receipt opt-in behind a server-side flag rather than a per-call param.

cat0825:
Author follow-up: the privacy feedback has been addressed in the current branch. Retrieval receipt hashes now use HMAC-SHA256 with a fresh per-receipt 32-byte salt that is never emitted, preserving duplicate detection only within a single receipt while preventing cross-receipt linkage or offline dictionary precomputation. I also added regression coverage for intra-receipt equality and cross-receipt non-linkability. This is ready for re-review.

PR #1033, opened 2026-06-01
fix(pipecat): handle missing profile in memory retrieval

### LINKED ISSUE #1027: BUG: AttributeError when profile is None in _retrieve_memories()

cat0825:
## Summary

Fixes a crash in `SupermemoryPipecatService._retrieve_memories()` when the Supermemory API returns `profile=None` for users with no stored memories.

Closes #1027.

## What changed

- Treats a missing profile as empty static/dynamic memories.
- Preserves existing search result handling.
- Adds a regression test for the no-profile response path.

## Validation

- `python -m pytest tests/test_service.py`


cat0825:
Author follow-up: current branch is ready for review. The Pipecat memory retrieval path now treats a missing profile as empty static/dynamic memories while preserving existing search result handling, with a regression test for the no-profile API response.

PR #1041, opened 2026-06-03
fix(pipecat-sdk): handle null profile object safely

SiddheshCodes4554:
This PR fixes a crash in Supermemory Pipecat memory retrieval when the API returns a null profile for a user with no stored memories. In that case, _retrieve_memories() previously accessed response.profile.static and response.profile.dynamic directly, which raised an AttributeError and caused the retrieval flow to fail.

**What changed**

1. Added a null check around the profile object returned by the Supermemory API.
2.  Safely fall back to empty lists when profile is None.
3. Kept search result handling defensive as well so missing search data does not break retrieval.
4.  Added a regression test covering the brand-new-user case where the API returns profile=None.

**Why this was needed**

When a user has no existing memories, the API can legitimately return profile=None. The previous implementation assumed profile always existed, which caused the Pipecat service to crash during memory retrieval and surface a MemoryRetrievalError to callers. This makes the empty-state path brittle and blocks first-time users from using the memory feature cleanly.

**Behavior after this change**

1. Users with no stored memories no longer trigger an AttributeError.
2. _retrieve_memories() now returns an empty profile structure instead of crashing.
3. The downstream context enhancement flow can continue safely even when there are no memories to inject.

**Verification**

1. Added a regression test for the null-profile path in packages/pipecat-sdk-python/tests/test_empty_profile.py.
2. Verified the test passes locally with unittest.

### DIFF
--- a/packages/pipecat-sdk-python/src/supermemory_pipecat/service.py
+++ b/packages/pipecat-sdk-python/src/supermemory_pipecat/service.py
@@ -146,14 +146,17 @@ class SupermemoryPipecatService(FrameProcessor):
 
             response = await self._supermemory_client.profile(**kwargs)
 
+            profile = getattr(response, "profile", None)
+            search_results_response = getattr(response, "search_results", None)
+
             search_results = []
-            if response.search_results and response.search_results.results:
-                search_results = response.search_results.results
+            if search_results_response and search_results_response.results:
+                search_results = search_results_response.results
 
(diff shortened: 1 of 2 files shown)


ishaanxgupta (approved):

ishaanxgupta:
@SiddheshCodes4554 Thanks for the contribution

PR #1048, opened 2026-06-04
Refine Codex and OpenCode OAuth integration status

ishaanxgupta:
## What changed

- Added /auth/agent-connect in the web app as an alias for the agent OAuth connect page and fixed the Codex icon on that flow.
- Updated integrations so plugin cards only show connected after the generated plugin key has actually been used by the plugin.
- Replaced the connected dropdown with a modal-based connection detail view.
- Removed backend-schema details like saved-note/conversation format and session IDs from plugin memory modals.
- Updated OpenCode setup copy to use the plugin browser login flow instead of manual API-key export, and added its GitHub link.

## Why

Generating a plugin key from the web UI is only setup, not a completed plugin connection. The card should not show Connected until the plugin/OAuth flow actually authenticates and uses the key.


ishaanxgupta:
image
image
image
image
image

**Opencode**
image

MaheshtheDev (approved):
LGTM

ThakerKush (approved):

PR #1053, opened 2026-06-05
feat(mcp): add privacy-safe recall receipts

SiddheshCodes4554:
This PR adds an optional, privacy-safe receipt mechanism for MCP recall operations so users can debug cross-client memory behavior without exposing raw memory content, prompts, or repository text.

**Why**

In multi-client workflows (Cursor, Claude Code, VS Code, etc.), debugging failures like:

 - “memory wasn’t recalled”
 - “wrong project context was retrieved”
 - “search worked but context wasn’t loaded”

currently requires sharing sensitive raw content.

**What’s Included**

 - Added a new receipt helper module for:
receipt shape/type
 - SHA-256-based privacy hashing (with optional salt)
 - similarity score bucketing
 - standardized receipt log formatting
 - Wired receipt generation into recall flow (both profile and non-profile paths)
 - Added optional server-side receipt logging via env flags
 - Documented receipt behavior and configuration in MCP README

### DIFF
--- a/apps/mcp/README.md
+++ b/apps/mcp/README.md
@@ -100,6 +100,7 @@ Search memories and get user profile.
 {
   "query": "What are the user's programming preferences?",
   "includeProfile": true,
+  "includeReceipt": false,
   "containerTag": "optional-project-tag"
 }
 ```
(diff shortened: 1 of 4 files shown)


ishaanxgupta (commented):
The receipt hashing is deterministic when `RECEIPT_HASH_SALT` is unset because `privacyHash()` defaults to `salt = ""` and hashes `${salt}:${input}` with plain SHA-256. That means raw queries, project ids, memory ids, and short memory contents can be dictionary-guessed and correlated across receipts, which contradicts the “privacy-safe” behavior described in the PR. Please either require a secret salt and fail/disable receipts without it, or use a per-receipt random HMAC salt that is not emitted so receipts are unlinkable by default.

MaheshtheDev:
not planned as of now

PR #1059, opened 2026-06-06
fix(web): strengthen URL validation and redirect handling in OG scraper

RinZ27:
Strengthened the `isPrivateHost` check to include the AWS/GCP/Azure metadata IP (`169.254.169.254`) and other loopback/unspecified addresses (`0.0.0.0`, `::`). This ensures that the OG scraper cannot be used to probe sensitive cloud metadata or internal services.

Switched the `fetch` call to `redirect: 'manual'` to prevent automated redirect bypasses. The implementation now manually validates the redirect location against the private host blocklist before following it, providing a more robust defense against SSRF attempts.

Added a `Content-Type` check to ensure the scraper only processes `text/html`. This prevents the server from attempting to download and process large binary files, reducing potential resource exhaustion during scraping operations.

Refactored the HTML metadata extraction into a reusable `processHtml` helper to maintain clean logic across the new redirect handling flow. All existing functionality for Open Graph and Twitter tags remains intact.

### DIFF
--- a/apps/web/app/api/og/route.ts
+++ b/apps/web/app/api/og/route.ts
@@ -19,7 +19,9 @@ function isPrivateHost(hostname: string): boolean {
 	if (
 		lowerHost === "localhost" ||
 		lowerHost === "127.0.0.1" ||
+		lowerHost === "0.0.0.0" ||
 		lowerHost === "::1" ||
+		lowerHost === "::" ||
 		lowerHost.startsWith("127.") ||
 		lowerHost.startsWith("0.0.0.0")
 	) {
@@ -30,6 +32,7 @@ function isPrivateHost(hostname: string): boolean {
 		/^10\./,
 		/^172\.(1[6-9]|2[0-9]|3[01])\./,
 		/^192\.168\./,
+		/^169\.254\./, // Link-local / Metadata service
 	]
 
 	return privateIpPatterns.some((pattern) => pattern.test(hostname))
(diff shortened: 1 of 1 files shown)


ishaanxgupta (commented):
The redirected fetch is no longer protected by the timeout. The code clears `timeoutId` immediately after the first response, then follows `redirectUrl` using the same `controller.signal`; at that point there is no timer left to abort the second request, so a slow/hanging redirected host can still tie up the route. Please keep a timeout active for the second fetch as well, for example by using a fresh controller/timeout around the redirect request or by moving `clearTimeout` after all fetches are complete.

RinZ27:
Nice catch @ishaanxgupta. The premature `clearTimeout` would've definitely been an issue if the redirect target hung. I decided to move the cleanup into a `finally` block that covers both the initial fetch and any potential redirects, keeping the `AbortController` active until the very end. Just pushed the update.

ishaanxgupta (approved):

apps/web/app/api/og/route.ts
```diff
@@ -19,7 +19,9 @@ function isPrivateHost(hostname: string): boolean {
 	if (
 		lowerHost === "localhost" ||
 		lowerHost === "127.0.0.1" ||
+		lowerHost === "0.0.0.0" ||
 		lowerHost === "::1" ||
+		lowerHost === "::" ||
```
MaheshtheDev, line 24: This IPv6 check — and the existing `=== "::1"` on line 23 — never matches. `URL.hostname` returns IPv6 literals **with brackets**, so the real values are `"[::]"` / `"[::1]"`:

```js
new URL("http://[::1]/").hostname // => "[::1]"
new URL("http://[::]/").hostname  // => "[::]"
```

So both equality checks are dead code and IPv6 loopback/unspecified hosts aren't actually blocked. Suggest stripping brackets before comparing (e.g. `const h = lowerHost.replace(/^\[|\]$/g, "")`), and ideally also covering `fc00::/7`, `fe80::/10`, and IPv4-mapped `::ffff:169.254.169.254`.

apps/web/app/api/og/route.ts
```diff
@@ -19,7 +19,9 @@ function isPrivateHost(hostname: string): boolean {
 	if (
 		lowerHost === "localhost" ||
 		lowerHost === "127.0.0.1" ||
+		lowerHost === "0.0.0.0" ||
```
MaheshtheDev, line 22: Redundant — `lowerHost.startsWith("0.0.0.0")` on line 26 already matches `"0.0.0.0"` exactly, so this line can be dropped.

apps/web/app/api/og/route.ts
```diff
@@ -179,61 +182,70 @@ export async function GET(request: Request) {
...
+					if (
+						!isValidUrl(redirectUrl) ||
+						isPrivateHost(new URL(redirectUrl).hostname)
+					) {
+						return Response.json(
+							{ error: "Invalid or private redirect URL" },
+							{ status: 400 },
+						)
+					}
+					// For simplicity, we only follow one level of redirect manually to re-validate IP
```
MaheshtheDev, line 208: Worth noting this follows only a single redirect hop — the second fetch uses `redirect: "error"`, so a legit chain (e.g. `http → https → canonical`) will throw and fall through to the 500 handler. Probably an acceptable tradeoff for the security win, but flagging in case multi-hop sites matter for this scraper.

MaheshtheDev (commented):
Overall this is a worthwhile hardening change and it fixes two real SSRF holes that exist in `main`:

- **Metadata range block (`169.254.0.0/16`)** — correct call. I verified it's robust against the usual numeric-encoding bypasses, since `new URL()` normalizes them *before* the regex runs: `http://2852039166/` (decimal), `http://0xA9FEA9FE/` (hex), and `http://0251.0376.0251.0376/` (octal) all parse to hostname `169.254.169.254`.
- **`redirect: "manual"` + re-validation** — important. The current default `redirect: "follow"` lets an attacker bypass `isPrivateHost` with a public URL that 30x-redirects to an internal IP, so this was a live vuln.

One fundamental gap to call out: every check here operates on the **hostname string** — DNS is never resolved. A domain whose A record points at `169.254.169.254` (or any private IP) bypasses all of this, including the redirect re-validation. Closing that requires resolving the host and validating the resolved IP (or an egress allowlist/proxy). Fine as a follow-up, but I'd avoid treating SSRF as fully closed until then.

A few smaller inline notes below.

PR #1063, opened 2026-06-08
fix(openai-sdk): use AsyncSupermemory and correct client API calls

icyzh:
The openai-sdk `SupermemoryTools` class initializes a sync `Supermemory()` client but all its methods are `async` and `await` it `TypeError` at runtime. On top of that, every `.add()` call targets the top-level client, but `AsyncSupermemory` exposes `add()` on the `.memories` sub-resource (`client.memories.add(...)`). Confirmed locally with the installed SDK:

    `client.add(...)` raises `AttributeError: 'AsyncSupermemory' object has no attribute 'add'`
    `client.memories.add(...)` works correctly
    `client.memories` exists, `client` does not have `add`

That missing method means every memory-add call silently hits the `except Exception` handler and returns `{"success": false, "error": "..."}`. Even if it did reach the API, the raw Pydantic response gets fed to `json.dumps()` without `.model_dump()`  another `TypeError`.

This rewrites the client init to use `AsyncSupermemory`, changes all `.add()` calls to `.memories.add()`, and calls `.model_dump()` on Pydantic responses before serialization. Test imports are also fixed  ,the first fallback referenced a non-existent package name (`supermemory_openai_sdk`) and the second used a wrong module path.

### DIFF
--- a/packages/openai-sdk-python/pyproject.toml
+++ b/packages/openai-sdk-python/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "hatchling.build"
 
 [project]
 name = "supermemory-openai-sdk"
-version = "1.0.3"
+version = "1.0.4"
 description = "Memory tools for OpenAI function calling with supermemory"
 readme = "README.md"
 license = "MIT"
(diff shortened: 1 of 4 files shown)


sohamd22 (approved):
Looks good to me!

ishaanxgupta:
@icyzh Please bump the package

icyzh:
Done , i have bumped the package

MaheshtheDev (approved):
reapproval from soham

PR #1073, opened 2026-06-09
Make integrations page public for guests

ved015:
## Summary
- Allow logged-out users to load `/?view=integrations` without being redirected to login
- Add a public integrations header with Supermemory branding and a login CTA
- Run IntegrationsView in guest mode with static catalog data only, no auth-backed queries
- Route all guest Connect actions to login with a redirect back to the integrations page

Screenshot 2026-06-09 at 9 32 33 PM

### DIFF
--- a/apps/web/app/(app)/page.tsx
+++ b/apps/web/app/(app)/page.tsx
@@ -582,6 +582,7 @@ export default function NewPage() {
 		viewMode === "dashboard" || (viewMode === "graph" && isMobile)
 	const isGraphMode = viewMode === "graph"
 	const showBottomNav = isMobile && !!session && !isChatView
+	const isPublicIntegrations = !session && viewMode === "integrations"
 
 	return (
 		<HotkeysProvider>
@@ -608,7 +609,9 @@ export default function NewPage() {
 						/>
 					</div>
 				)}
-				{!session && viewMode === "mcp" ? (
+				{isPublicIntegrations ? (
+					<PublicHeader variant="integrations" />
+				) : !session && viewMode === "mcp" ? (
 					<PublicHeader />
 				) : (
 					<Header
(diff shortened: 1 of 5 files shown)


MaheshtheDev:
### Review notes

**🟡 Main one — `EnsureWorkspace` now skips the onboarding guard for logged-in users too**

`apps/web/components/ensure-workspace.tsx` — broadening `isMcpPublicPage` → `isPublicAppPage` makes `if (isPublicAppPage) return` bail for *everyone* on `/?view=integrations`, not just guests. A logged-in user without an org no longer gets redirected to `/onboarding`; they fall through to a half-populated authenticated integrations page (`publicMode=false` but `org?.id` null).

Suggest scoping the bypass to guests:
```ts
if (isPublicAppPage && !session) return   // guests stay public
// logged-in users still get the org / onboarding guarantee
```
(and add the same `!session` qualifier to `showLoading`).

**🟢 Minor (non-blocking)**
- Middleware: the new integrations bypass correctly requires `pathname === "/"`, but the older `view=mcp` check doesn't — worth matching someday.
- `redirectToLogin` uses `window.location.href`, so stray params (`cat=`, `add=`…) ride into the post-login redirect; the header CTA hardcodes the clean `/?view=integrations`.
- Guests: `isItemConnected` returns false, so the "Connected" filter chip shows a `0` count instead of being hidden (cosmetic).

Rest looks good — `AutumnProvider` is mounted unconditionally so disabled `useCustomer` is safe, and all guest-visible queries/mutations are properly gated on `!publicMode`.

MaheshtheDev:
Re-reviewed `f62c6ecad` — the `isGuestPublicAppPage = isPublicAppPage && !session` split fixes the main issue: logged-in orgless users now hit the `/onboarding` redirect again, while guests stay on the public page. The added `pathname === "/"` guard is a nice bonus that lines up with the middleware check. 👍

Minor/optional: during `isRestoring`, an actually-logged-in user has `session` momentarily null, so they'll see a brief flash of the guest view before it flips to authenticated. Cosmetic, matches existing mcp behavior — not blocking.

MaheshtheDev (approved):

PR #1082, opened 2026-06-10
fix(deps): bump next to 16.0.7 in memory-graph-playground (critical CVE)

aaronjmars:
## Summary

Bumps `next` from **16.0.3** to **16.0.7** in `apps/memory-graph-playground` to resolve [GHSA-9qr9-h5gf-34mp](https://github.com/advisories/GHSA-9qr9-h5gf-34mp) — a **critical RCE in the React flight protocol** affecting `>= 14.3.0-canary.77 < 16.0.7` (16.0.x line patched in 16.0.7).

This surfaced via the Socket Security scan on #1046, but the vulnerable version is on `main` itself (that PR doesn't touch dependencies), so fixing it separately here.

## Changes

- `apps/memory-graph-playground/package.json`: `next` `16.0.3` → `16.0.7`
- `bun.lock`: regenerated with **bun 1.3.6** (matching CI) via `bun install --lockfile-only` — only the `next`/`@next/*` 16.0.x entries change, no other dependencies touched

`apps/web` already pins `^16.0.11` and the chatapp test fixture is excluded from the workspace, so the playground was the only vulnerable entry in the lockfile.

### DIFF
--- a/apps/memory-graph-playground/package.json
+++ b/apps/memory-graph-playground/package.json
@@ -11,7 +11,7 @@
   },
   "dependencies": {
     "@supermemory/memory-graph": "workspace:*",
-    "next": "16.0.3",
+    "next": "16.0.7",
     "react": "19.2.0",
     "react-dom": "19.2.0"
   },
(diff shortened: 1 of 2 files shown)


MaheshtheDev (approved):
LGTM

PR #1090, opened 2026-06-10
feat: add listMemories tool to MCP server

Nemskri:
### What does this PR do?
This PR implements the requested `listMemories` tool for the Supermemory Model Context Protocol (MCP) server, addressing **Issue #1030**. This allows AI assistants (like Claude) to enumerate and audit stored memories in a paginated/sortable list instead of only retrieving them via semantic `recall`.

---

### Implementation Details

1. **MCP Tool Registration (`apps/mcp/src/server.ts`)**:
   - Registered the `listMemories` tool.
   - Defined Zod schemas for `limit` (max 200), `cursor` (opaque token for pages), `sort` (`created_desc`, `created_asc`, `updated_desc`), `filter` (optional search substring), and optional project `containerTag`.

2. **SDK client helper (`apps/mcp/src/client.ts`)**:
   - Added `listMemories` helper method to the `SupermemoryClient`.
   - Queries `this.client.documents.list` under the hood (per the SDK v4 design where `/v3/documents/list` returns the list of memories).
   - Applies local case-insensitive substring `filter` matching on titles and text.

3. **Standard Transport Fix (`apps/mcp/src/index.ts`)**:
   - Set the default server transport to `"sse"` in `SupermemoryMCP.serve(...)`.
   - This resolves a `400 Bad Request` connection error when standard SSE clients (like the official MCP Inspector) try to connect without a pre-initialized session ID header.

---

### Verification & Testing
- **Type Checking**: Verified that the type check passes on the `mcp` app target with `bunx tsc --noEmit --project apps/mcp/tsconfig.json`.
- **Formatting**: Ran `npx biome check` on modified files to ensure strict compliance with code formatting/linting rules (all checks passed).
- **Manual verification**: Connected successfully via the official **MCP Inspector** using standard SSE at `http://localhost:8788/mcp` and verified that the schema parameters show up correctly and run under local mock authentications.

### DIFF
--- a/apps/mcp/src/client.ts
+++ b/apps/mcp/src/client.ts
@@ -255,6 +255,61 @@ export class SupermemoryClient {
 		}
 	}
 
+	// List all memories using SDK
+	async listMemories(
+		containerTag?: string,
+		page = 1,
+		limit = 50,
+		sort: "createdAt" | "updatedAt" = "createdAt",
+		order: "asc" | "desc" = "desc",
+		filter?: string,
+	): Promise<{ memories: Memory[]; nextCursor?: string }> {
+		try {
+			const result = await this.client.documents.list({
+				containerTags: containerTag ? [containerTag] : [this.containerTag],
+				page,
+				limit,
+				sort,
+				order,
+				includeContent: true,
+			})
+
+			let memories: Memory[] = result.memories.map((r) => {
+				const text = limitByChars(r.content || r.summary || "")
+				return {
+					id: r.id,
+					similarity: 1.0, // Not applicable for a raw list
+					title: r.title || undefined,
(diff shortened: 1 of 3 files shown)


ishaanxgupta:
Hi @Nemskri, thanks for the contribution
Right now listMemories calls the documents list API with includeContent: true, which means the tool returns full document bodies. In my test this caused Claude to hit output limits even with limit: 5, because one document had ~48k chars and the response included that same text in both content and memory.
Either we need to use a new endpoint for listing memories as the tool name is list-memories or give a flag like includeContent default as false to add less content and only add fields like summary,createdat etc in the exisying endpoint

I will have an internal discussion and continue it from here.

Issue #1093, opened 2026-06-11
Self-hosted server: unbounded memory growth in local embedding workers under continuous ingestion (reaches 22–50 GB)


chetanunadkat:
## Summary

On the self-hosted binary (`v0.0.2`, macOS arm64), `supermemory-server` grows to **20–50 GB RSS** during sustained ingestion and never releases it. On a 16 GB machine this exhausts RAM + swap. The growth is in the local embedding workers (onnxruntime-web / transformers.js WASM), and the only memory-reclamation path in the pool can't fire while the server is busy.

## Environment

- supermemory-server `v0.0.2` (self-hosted, `bun build --compile` binary), macOS 15 / arm64
- Embeddings: local, default model `Xenova/bge-base-en-v1.5`
- Config: `SUPERMEMORY_LOCAL_EMBEDDING_POOL_SIZE=3`, threads/worker = 1 (default), `SUPERMEMORY_LOCAL_EMBEDDING_IDLE_TIMEOUT_MS=120000` (default)
- Continuous ingestion (a few documents/min) via the documents API

## Observed

- `footprint` of the process: **~22 GB**, of which **~21 GB is dirty "WebKit malloc"** (the WASM `ArrayBuffer`s + JS heap), climbing ~3–4 GB/min while ingesting; peak observed ~50 GB before it was killed.
- `sample` shows the hot threads inside ort-web WASM inference even between requests.
- Document queue was small/drained throughout — this is steady-state ingestion, not a backlog.
- Restarting with `SUPERMEMORY_LOCAL_EMBEDDING_POOL_SIZE=1` + `SUPERMEMORY_SKIP_EMBEDDING_PREWARM=1` keeps it flat at ~1.2 GB **until embedding load resumes** (so the workaround reduces blast radius / defers, it doesn't remove the leak).

## Root cause (from inspecting the bundled embedding worker in the v0.0.2 binary)

Two things combine:

[...]

PR #1094, opened 2026-06-11
docs(self-hosting): document local embedding memory behavior

chetanunadkat:
## What

Adds a **Memory considerations** subsection to the self-hosting embedding docs explaining that each local embedding worker's WebAssembly heap only grows (up to ~4 GB) and is reclaimed only when the pool goes fully idle.

## Why

Self-hosters on small hosts hit unbounded RSS growth (20–50 GB) under continuous ingestion because peak memory scales as roughly `POOL_SIZE × ~4 GB` and the only reclamation path (idle shutdown) never fires while the server is busy. The current docs present `POOL_SIZE` purely as a throughput knob, with no memory-cost guidance. This adds the missing budgeting note and the safe-default recommendation for ≤16 GB hosts.

Documents the user-facing/operational side of #1093. The worker/pool implementation lives in the closed binary, so this PR only covers the docs; the issue proposes the code-level fix (recycle busy workers after N batches / a memory ceiling).

## Change

One subsection added to `apps/docs/self-hosting/configuration.mdx`, right after the embedding-performance table. Docs-only, no code changes.


ishaanxgupta:
@chetanunadkat I see some conflicts here

ishaanxgupta:
@chetanunadkat please fix the conflicts

PR #979, opened 2026-05-20
feat(web): overhaul integrations page with featured hero, category rails and MCP setup modal

MaheshtheDev:
- Rebuild the integrations catalog as horizontally-scrollable category rails
  (MCP, Plugins, Knowledge bases, Apps & extensions) with header chevrons
- Add an auto-rotating Featured hero with problem-led headlines, backdrop
  imagery and a gradient FEATURED badge
- Surface all 9 MCP clients as individual cards that open a setup modal
  styled like the plugin connect flow (reuses MCPSteps embedded variant)
- Align MCP setup steps and the Claude Desktop timeline to the design system
- Merge the Integrations and Connections settings tabs into one entry

PR #1130, opened 2026-06-17
feat(web): show import status on X bookmarks integration card

sreedharsreeram:
## What

The X bookmarks ("Import X bookmarks") card on the integrations page never reflected any state, even after importing tweets — unlike connectors and plugins, which show "Connected"/"Active".

After this change, once you've imported at least one tweet the card shows an **"Imported · {last import time}"** pill (mirroring the plugin "Active · {time}" style) and is included in the **Connected** filter.

Since importing X bookmarks is a one-time/occasional action rather than a live connection, it intentionally says **"Imported"** with the last-import timestamp, not "Connected".

## Why

The page's status logic (`isItemConnected` + `renderStatus`) only handled the `plugin` and `connector` item kinds. The X bookmarks card is an `import` kind, so it always rendered just "Connect". This was a regression from #979 (the integrations overhaul), which rebuilt the page around item kinds and dropped the previous "{N} tweets imported" indicator.

## How

- Add a single documents query (`@post/documents/documents`, `categories: ["tweet"]`, `limit: 1`, newest first) that yields both the org-wide tweet count (`pagination.totalItems`) and the latest tweet's `createdAt`.
- `isItemConnected` returns `true` for `import` when the count is `> 0`.
- `renderStatus` renders a new `ImportedPill` ("Imported · {relative time}") for `import` when there's ≥1 tweet.

## Testing

- `tsc` and Biome pass; no new errors introduced by this change.
- Verified the status source of truth against a local backend by seeding `type='tweet'` documents — the count/latest-timestamp the pill renders reflect them.

> Note: `tsc` reports two **pre-existing**, unrelated `granola` errors in this file (`CONNECTOR_META` + an icon record missing the `granola` provider). They exist on `main` and are not touched by this PR.

### DIFF
--- a/apps/web/components/integrations-view.tsx
+++ b/apps/web/components/integrations-view.tsx
@@ -1059,6 +1059,29 @@ function ConnectionsCountPill({ count }: { count: number }) {
 	)
 }
 
+function ImportedPill({
+	lastImportedAt,
+}: {
+	lastImportedAt: string | number | Date | null
+}) {
+	return (
+		<span
+			className={cn(
+				dmSans125ClassName(),
+				"flex shrink-0 items-center gap-1.5 text-[12px] font-medium text-[#00AC3F] sm:text-[13px]",
+			)}
+		>
+			<span className="size-[7px] rounded-full bg-[#00AC3F]" />
+			Imported
+			{lastImportedAt && (
+				<span className="text-[11px] font-normal text-[#737373]">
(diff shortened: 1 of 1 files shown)


MaheshtheDev (changes requested):
need to tweak this and kinda show imported with last time it connected

MaheshtheDev (approved):

PR #1135, opened 2026-06-18
Show MCP connection status in integrations

ishaanxgupta:
## Summary
- Parse enabled MCP OAuth API keys from the integrations page key list.
- Show MCP as connected in MCP integration cards, the Active filter count, featured CTA, and active connections rail.
- Reuse the existing MCP metadata signal used elsewhere in the app (`sm_source: "mcp"` or `sm_kind: "mcp_oauth_exchange"`).

image

### DIFF
--- a/apps/web/components/integrations-view.tsx
+++ b/apps/web/components/integrations-view.tsx
@@ -93,6 +93,19 @@ interface ConnectedKey {
 	createdAt?: string | null
 }
 
+interface ConnectedMcpKey {
+	keyId: string
+	keyStart: string | null
+	lastRequest?: string | null
+	createdAt?: string | null
+}
+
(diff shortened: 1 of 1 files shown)


ishaanxgupta:
Update after latest commit:

- MCP connection state is now shown once in the MCP section header, before the carousel arrows, because the API does not return per-client MCP identity.
- MCP client cards now use compact `+` actions instead of text `Connect` buttons.
- The previous per-card MCP connected indicator was removed so all MCP clients are not shown as individually connected.

MaheshtheDev (changes requested):
add mcp icon instead of black icon in active status

MaheshtheDev (approved):

PR #1213, opened 2026-07-10
fix(web): bypass auth proxy for local dev

GautamSharma99:
Description
  This removes a manual setup step for OSS contributors by making the web middleware
  automatically allow plain localhost development.

  Previously, CONTRIBUTING.md instructed contributors to manually edit the web proxy/middleware
  before running bun run dev:local. Without that edit, local requests could be redirected
  through the auth proxy instead of reaching the local app directly.

  Changes

  - Add a local-dev host bypass in apps/web/middleware.ts
      - localhost
      - 127.0.0.1
      - ::1

  - Run the bypass before reading Better Auth session cookies
  - Remove the obsolete manual proxy-edit step from CONTRIBUTING.md

  Why
  bun run dev:local should work without contributors modifying tracked source files. This keeps
  local setup simpler and avoids accidental dirty worktrees.

  Testing

  - bunx biome check CONTRIBUTING.md apps/web/middleware.ts

---
**Session Details**
- Session: [View Session](https://supermemory.us1.vorflux.com/agent-sessions/325a3f70-5370-43ea-830f-c05a917a5c7f)
- Requested by: Unknown
- Address comments on this PR. Add `(aside)` to your comment to have me ignore it.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -36,16 +36,7 @@ Before you begin, ensure you have the following installed:
    # You'll need to add your API keys and database URLs
    ```
 
-4. **Change proxy for local development**
-
-   Add this in your `proxy.ts`(apps/web) before retrieving the cookie (`getSessionCookie(request)`):
-
-   ```ts
-   if (url.hostname === "localhost") {
-     return NextResponse.next();
-   }
-
-5. **Start the Development Server**
+4. **Start the Development Server**
 
    ```bash
    bun run dev:local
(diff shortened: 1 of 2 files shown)


Dhravya:
Pushed two commits (maintainer edits): a merge with main, and a guard change — the localhost bypass now also requires `process.env.NODE_ENV === "development"`.

Reason: `getPublicRequestUrl` trusts the `x-forwarded-host` header, and this middleware 401-gates `/api/*`. Without the env guard, a production request with `x-forwarded-host: localhost` would skip that gate. NODE_ENV is inlined at build time, so the bypass is dead code in production bundles while still removing the manual proxy-edit step for local contributors.

PR #1233, opened 2026-07-11
Fix memory tooling regressions

### LINKED ISSUE #1184: Plugin transcript parser drops every line after the first in multi-line messages
## Summary

`parseTranscriptMessages` in `apps/web/lib/plugin-document.ts` builds its message-splitting regex with the `m` flag, and the terminating lookahead ends with `|$`:

```ts
const regex = new RegExp(
	`^\\s*(\\d+)\\.\\s+\\[(${TRANSCRIPT_ROLE_PATTERN})\\]\\s*([\\s\\S]*?)(?=^\\s*\\d+\\.\\s+\\[(?:${TRANSCRIPT_ROLE_PATTERN})\\]\\s*|$)`,
	"gm",
)
```

With `m`, `$` matches at **every line end**, not just end of input, so the lazy `([\s\S]*?)` body — clearly written to span lines — stops at the first newline. Every line of a message after the first is silently dropped.

## Reproduction

```js
const content = [
	"1. [user] Hello there",
	"Here is more context on line two",
	"memory id: mem_123",
	"2. [assistant] Sure!",
	"Second line of the reply",
].join("\n")
```

Actual parsed messages: `"Hello there"` and `"Sure!"`. The continuation lines appear in **no** message.

## Impact

- Codex session and Amp thread documents (`parseSessionTranscript`) render truncated conversations in the document modal — only the first line of each turn survives.
- `memory id: ...` lines inside a message body are dropped instead of being surfaced as artifacts by `extractArtifacts`, so the Memory ID chip never shows for multi-line turns.
- Session previews and user/assistant message counts are computed from truncated text.

This is amplified by `normalizeContent`, which converts literal `\n` escapes into real newlines *before* parsing — so multi-line bodies are the norm for these payloads, not the edge case.

## Expected behavior

Message bodies should span lines until the next `N. [role]` header or the true end of input. The last message in a transcript should keep all of its lines.

## Suggested fix

Replace the `|$` alternative in the lookahead with a real end-of-input assertion that isn't affected by the `m` flag: `|(?![\s\S])`. One-line change; the adjacent `parseClaudeCodeTurns` regex is unaffected (it uses `$` without the `m` flag, where it already means end of input).

### LINKED ISSUE #1191: fix(ai-sdk): use z.coerce.number() for limit/offset params to handle string inputs from LLMs
## Description

When using the AI SDK tools from `@supermemory/tools`, calls to `searchMemories` and `documentList` fail with a validation error when the LLM generates `limit` or `offset` as a string instead of a number:

```
tool call validation failed: parameters for tool searchMemories did not match schema: errors: [`/limit`: expected number, but got string]
```

## Root Cause

The zod schemas for `searchMemories` and `documentList` define `limit` and `offset` as `z.number()`. However, LLMs occasionally emit numeric parameters as JSON strings (e.g., `"10"` instead of `10`). `z.number()` rejects strings, while `z.coerce.number()` would silently coerce them.

## Affected Fields

| Tool | Field | Current Schema |
|---|---|---|
| `searchMemories` | `limit` | `z.number().optional().default(10)` |
| `documentList` | `limit` | `z.number().optional().default(10)` |
| `documentList` | `offset` | `z.number().optional()` |

## Fix

Change `z.number()` → `z.coerce.number()` in `packages/tools/src/ai-sdk.ts` for all three fields.

`z.coerce.number()` is a standard Zod v4 pattern that converts string inputs to numbers during validation. If the value is already a number, it passes through unchanged. If it is `undefined` (for optional fields), it remains undefined. Only string-form numbers get coerced.

## Relevant Code

https://github.com/supermemoryai/supermemory/blob/main/packages/tools/src/ai-sdk.ts

## Environment

- `@supermemory/tools`: 2.0.0
- `zod`: ^4.1.5
- Used via `@supermemory/tools/ai-sdk` export

### LINKED ISSUE #1205: Dead code in search threshold schemas: unreachable `v === undefined` checks and no-op `.transform(Number)`
## Summary

In `packages/validation/api.ts`, three threshold fields use a `.refine()` that
checks `v === undefined`, but this branch is unreachable, and a `.transform(Number)`
that is a no-op.

Affected fields:
- `SearchRequestSchema.chunkThreshold`
- `SearchRequestSchema.documentThreshold`
- `Searchv4RequestSchema.threshold`

## Why the code is dead

Each field is defined as:

    z.number().optional().default(0).refine((v) => v === undefined || (v >= 0 && v <= 1), {...}).transform(Number)

Zod applies `.default()` before downstream processing. When the field is omitted,
the parsed input is `undefined`, `.default(0)` replaces it with `0`, and only then
does `.refine()` run. So `v` is always a number, never `undefined`, the
`v === undefined ||` branch can never execute.

Additionally, `.transform(Number)` runs on a value that is already `z.number()`,
so it is a no-op.

## Question before I open a PR

Happy to submit a PR removing the dead `v === undefined` checks and the redundant
`.transform(Number)` across all three fields. Before I do:

Is the `.default(0)` intentional? If "unset" is meant to be distinguishable from
an explicit `0`, that may be worth addressing separately, otherwise I'll keep the
cleanup purely to the dead code and leave the defaults as-is.

### LINKED ISSUE #1206: Vercel AI SDK middleware silently drops tool-call and tool-result messages when saving conversations
## Summary

In `packages/tools/src/vercel/middleware.ts`, `convertToConversationMessages`
(lines 18–76) only maps `text` and `file` content parts. Any other part type,
including `tool-call` (assistant) and `tool-result` (tool role), hits `return null`
and is removed by `.filter((part) => part !== null)`.

Because a pure tool-call assistant turn has *only* tool-call parts, all its parts
become `null`, `contentParts.length` is `0`, and the entire message is skipped.
The same happens to tool-role messages carrying `tool-result`.

Result: conversations that use tool calling via the Vercel middleware are saved to
Supermemory with all tool interactions stripped. The tool that was called, its
arguments, and its result are lost from the stored conversation.

## Evidence

- `convertToConversationMessages` handles only `c.type === "text"` and
  `c.type === "file"`; everything else returns `null` (map at ~line 35–50).
- Searching the vercel folder for `tool-call` / `tool-result` / `toolCallId`
  returns nothing, there is no tool handling at all.
- The OpenAI middleware (`openai/middleware.ts`) does preserve tool metadata, so
  Vercel-based memories are inconsistent with OpenAI-based ones.
- `ConversationMessage` (`conversations-client.ts`) already supports
  `tool_calls?: ToolCall[]` and `tool_call_id?: string`, and the
  `/v4/conversations` endpoint doc states it "supports tool calls and responses",
  the receiving side is ready; the Vercel adapter just never sends them.

## SDK version

`packages/tools/package.json` pins `ai: ^5.0.29` (AI SDK v5), where tool parts are:
- assistant: `{ type: "tool-call", toolCallId, toolName, input }`
- tool:      `{ type: "tool-result", toolCallId, toolName, output }`

## Proposed fix

Extend the content-part mapping in `convertToConversationMessages` to handle v5
tool parts:
- `tool-call` → append to the message's `tool_calls` as
  `{ id: toolCallId, type: "function", function: { name: toolName, arguments: JSON.stringify(input) } }`
- `tool-result` → emit a `tool`-role message with `tool_call_id: toolCallId` and
  the serialized `output` as content

Happy to submit a PR with unit tests covering a tool-call + tool-result round trip.
Want me to proceed?

Rohithmatham12:
## Summary
- coerce AI SDK memory tool numeric inputs so stringified limit/offset values validate
- preserve Vercel AI SDK tool-call and tool-result messages when saving conversations
- keep multi-line plugin transcript messages intact and remove dead threshold schema branches

## Fixes
- Fixes #1191
- Fixes #1205
- Fixes #1184
- Fixes #1206

## Validation
- vitest run packages/tools/src/ai-sdk.test.ts packages/tools/test/with-supermemory/unit.test.ts
- bun test apps/web/lib/plugin-document.test.ts
- biome check apps/web/lib/plugin-document.ts apps/web/lib/plugin-document.test.ts packages/tools/src/ai-sdk.ts packages/tools/src/ai-sdk.test.ts packages/tools/src/vercel/middleware.ts packages/tools/test/with-supermemory/unit.test.ts packages/validation/api.ts


ved015:
@Rohithmatham12 can you please resolve conflicts and remove the test files and mark it as ready for review

PR #1236, opened 2026-07-11
fix(openai-sdk): cap supermemory to <3.5 so a fresh install imports

### LINKED ISSUE #1235: supermemory-openai-sdk fails to import with the latest supermemory (3.50)
Hit this trying to use the openai SDK package. A fresh install pulls supermemory 3.50.0 (pyproject just says supermemory>=3.1.0) and then importing it blows up right away:

pip install supermemory-openai-sdk
python -c "import supermemory_openai"

Looks like supermemory 3.x renamed a bunch of stuff. MemoryAddResponse and MemoryGetResponse aren't in supermemory.types anymore (it's AddResponse, DocumentGetResponse etc now), and tools.py imports them at module top so the whole package won't import.

Even after fixing the imports, add_memory() calls client.memories.add(...) which is also gone in 3.50. client.memories only has forget and update_memory now, and add moved to the top level (client.add(...)).

It works inside the repo because uv.lock pins an older supermemory, but anyone doing a plain pip install from PyPI hits the import error immediately.

Happy to send a PR for this. I'd swap the imports to the new names and point add_memory at the new API, or just cap the supermemory version if you'd rather do the migration yourselves. MemoryObject is exported in __init__.py so I didn't want to touch that without checking. Let me know which way you want it and I'll take it.

Cintu07:
Fixes #1235

A fresh pip install supermemory-openai-sdk pulls the latest supermemory (3.50) and the package fails to import, because supermemory 3.5+ renamed the types it imports (MemoryAddResponse etc) and moved memories.add off client.memories.

The package's uv.lock already pins supermemory 3.4.0, which is what it actually works against, so this just caps the declared dependency to the same 3.4.x line. Confirmed a fresh install with the cap imports fine.

This is the minimal fix to unbreak installs. Happy to instead do the full migration to the newer 3.x API (client.add, AddResponse, etc.) if you'd rather go that route, just say the word

### DIFF
--- a/packages/openai-sdk-python/pyproject.toml
+++ b/packages/openai-sdk-python/pyproject.toml
@@ -26,7 +26,7 @@ classifiers = [
 requires-python = ">=3.8.1"
 dependencies = [
     "openai>=1.102.0",
-    "supermemory>=3.1.0",
+    "supermemory>=3.1.0,<3.5.0",
     "typing-extensions>=4.0.0",
     "requests>=2.25.0",
 ]
--- a/packages/openai-sdk-python/pyproject.toml
+++ b/packages/openai-sdk-python/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "hatchling.build"
 
 [project]
 name = "supermemory-openai-sdk"
-version = "1.0.4"
+version = "1.0.5"
 description = "Memory tools for OpenAI function calling with supermemory"
 readme = "README.md"
 license = "MIT"


Cintu07:
still broken on pypi as of today. supermemory 3.51.0 shipped yesterday and
supermemory-openai-sdk 1.0.4 still declares supermemory>=3.1.0, so a fresh
install pulls 3.51 and the import fails.

to be clear, this pr needs no decision from you. it caps to the same 3.4.x
line uv.lock already pins, one line, matches what the package actually works
against today. checks are green.

the full migration to the newer api is worth doing, but it's a bigger change
and shouldn't block unbreaking installs. i'll open that as a separate pr once
this lands.

@Dhravya mind taking a look? small one.

ishaanxgupta (approved):

ishaanxgupta:
Looks good. Please bump supermemory-openai-sdk from 1.0.4 to 1.0.5 before merging, since 1.0.4 is already published on PyPI and cannot be replaced

Cintu07:
> Looks good. Please bump supermemory-openai-sdk from 1.0.4 to 1.0.5 before merging, since 1.0.4 is already published on PyPI and cannot be replaced

done, bumped to 1.0.5 in 3f34afd. good to merge whenever you get a sec ^-^

Cintu07:
the 1.0.5 bump staled the approval, mind re-approving when you get a sec? this one is the quick cap. i also opened the full migration to the 3.x api as the proper fix in #1427, so take whichever you prefer

ved015 (approved):

ishaanxgupta (approved):

Issue #1237, opened 2026-07-11
Self-hosted v0.0.5: API key auth broken — 401 "Either userId or orgId not found" even with banner key


dupenodi:
## Environment

- **OS**: macOS arm64 (darwin-arm64)
- **supermemory-server**: upgraded `v0.0.3 → v0.0.5` via `supermemory-server upgrade`
- **Bun**: v1.3.4
- **Data dir**: custom `SUPERMEMORY_DATA_DIR` (project-local `.supermemory/`)
- **LLM**: `GEMINI_API_KEY` in `~/.supermemory/env` (unrelated to this bug)

## Steps to reproduce

1. Start from a working self-hosted install on **v0.0.3** (auth works: `POST /v3/documents/list` with the banner API key returns 200).
2. Run `supermemory-server upgrade` (lands on **v0.0.5**).
3. Restart the server against the same `SUPERMEMORY_DATA_DIR`.
4. Copy the `api key` from the startup banner (or from `$SUPERMEMORY_DATA_DIR/api-key`).
5. Call an authenticated route, e.g.:

```bash
curl -sS -X POST http://localhost:6767/v3/documents/list \
  -H "Authorization: Bearer <banner-api-key>" \
  -H "Content-Type: application/json" \
  -d '{}'
```

## Expected behavior

Same as v0.0.3 and the self-hosting docs: the API key printed at boot (stored in `$SUPERMEMORY_DATA_DIR/api-key`) authenticates Bearer requests. `POST /v3/documents/list` returns 200.

## Actual behavior

Every authenticated API call returns **401**:

```json
{"error":"Unauthorized","details":"Either userId or orgId not found. Are you using the right API key?"}
```

The server is unusable post-upgrade. The startup banner still prints the key and still says it is “auto-applied for unauthenticated localhost requests.”

## Error messages / logs

v0.0.5 also prints this on startup (not present on v0.0.3):

```
[better-auth/magic-link] `allowedAttempts` is ignored: tokens are consumed
atomically on the first verification call (GHSA-hc7v-rggr-4hvx). Any value
other than `1` has no effect; remove the option to silence this warning.
```

Note: better-auth / this error string already exist inside the v0.0.3 binary, so this may be a dependency-bump warning rather than a brand-new auth stack. The functional regression is real either way.

## Additional context (ruled out)

1. **Wrong API key** — compared programmatically: `.env` value, on-disk `api-key`, and startup banner key all match exactly.
[...]

PR #1238, opened 2026-07-11
docs: document pinned install for supermemory-server

dupenodi:
## What

Documents the self-hosted installer's existing version-pin argument in the Quickstart.

## Why

`install.sh` already accepts an explicit version (`$1`, e.g. `0.0.3`), but the Quickstart only showed `curl … | bash` (latest). After hitting a v0.0.5 auth regression (#1237), the pin/rollback path was the recovery route — it should be discoverable from the docs. The binary also nags “update available” on startup; when you intentionally pin, that message is safe to ignore.

## Changes

- New “Pin or change versions” subsection under Install in `apps/docs/self-hosting/quickstart.mdx`
- Example: `curl -fsSL https://supermemory.ai/install | bash -s -- 0.0.3`
- Link to GitHub release tags (`server-v<version>`)
- Note `supermemory-server upgrade` and the startup update nag

## Breaking changes

None.

## Related

- Closes nothing (docs only)
- Context: #1237

## Test plan

- [ ] Preview Self-Hosting → Quickstart; confirm “Pin or change versions” renders
- [ ] Confirm example matches `install.sh` (`TARGET="${1:-latest}"`)

### DIFF
--- a/apps/docs/self-hosting/quickstart.mdx
+++ b/apps/docs/self-hosting/quickstart.mdx
@@ -27,6 +27,24 @@ bunx supermemory local
 
 The installer detects your OS and architecture, downloads the right binary, verifies it, and (when run interactively) prompts you for an LLM API key. Supported platforms: macOS (Apple Silicon & Intel), Linux (x64 & arm64).
 
+### Pin or change versions
+
+Pass an explicit version to install (or roll back to) a specific release instead of `latest`:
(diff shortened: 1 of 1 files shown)


ved015 (approved):
LGTM !

ishaanxgupta (approved):

PR #1244, opened 2026-07-12
fix: accept a Supermemory API key programmatically in the OpenAI integration

### LINKED ISSUE #1242: OpenAI middleware should support programmatic API key configuration like other integrations
### Summary

The OpenAI integration is the only Supermemory integration that does not allow the API key to be supplied programmatically. Unlike the Vercel, Mastra, and Voltagent integrations, it only reads `SUPERMEMORY_API_KEY` from the environment, making it unusable in environments where the key is managed at runtime.

### Affected code

* `packages/tools/src/openai/middleware.ts:109`
* `packages/tools/src/openai/middleware.ts:425`
* `packages/tools/src/openai/index.ts:66`

### Current behavior

The OpenAI middleware reads the API key directly from:

```ts
process.env.SUPERMEMORY_API_KEY
```

and `OpenAIMiddlewareOptions` does not expose an `apiKey` option.

As a result, the middleware throws if the key is not present as an environment variable, even when the caller has the key available programmatically.

### Comparison

Other integrations (Vercel, Mastra, and Voltagent) support both patterns:

```ts
options.apiKey ?? process.env.SUPERMEMORY_API_KEY
```

and pass the result through the shared `validateApiKey` helper (`shared/context.ts`), providing consistent behavior across integrations.

### Why this matters

Many deployment environments provide secrets programmatically rather than through process-level environment variables. Requiring `SUPERMEMORY_API_KEY` to exist in `process.env` makes the OpenAI middleware less flexible than the other integrations and creates unnecessary API inconsistencies.

### Suggested fix

Add an optional `apiKey` field to `OpenAIMiddlewareOptions` and route API key resolution through the existing shared `validateApiKey` helper, matching the behavior of the Vercel, Mastra, and Voltagent integrations.

This would make all integrations consistent while reusing the existing API key validation logic.

rajarshidattapy:
## Summary

Fixes #1242.

The OpenAI integration was the only one that couldn't accept an API key programmatically.

The Vercel, Mastra, and Voltagent integrations all resolve:

```ts
options.apiKey ?? process.env.SUPERMEMORY_API_KEY
```

through the shared `validateApiKey` helper (`src/shared/context.ts`).

The OpenAI middleware instead read `process.env.SUPERMEMORY_API_KEY` directly in two places and exposed no `apiKey` field on `OpenAIMiddlewareOptions`.

As a result, `withSupermemory()` hard-threw whenever the key wasn't available as a process environment variable (for example when using a secrets manager, an edge runtime, or a multi-tenant server with a per-request key), with no way to supply one programmatically.

## Fix

This PR adds an optional `apiKey` field and routes all API key resolution through the shared `validateApiKey()` helper, bringing the OpenAI integration in line with the others.

### `src/openai/middleware.ts`

- Adds `apiKey?: string` to `OpenAIMiddlewareOptions`
- Resolves the key once via:

```ts
validateApiKey(options?.apiKey)
```

- Threads the resolved key through:
  - `supermemoryProfileSearch()` (now accepts an `apiKey` parameter instead of reading `process.env` when constructing the `Authorization` header)
  - `addSystemPrompt()`
- Passes the resolved key into `addMemoryTool()` instead of `process.env.SUPERMEMORY_API_KEY`

### `src/openai/index.ts`

- Replaces:

```ts
if (!process.env.SUPERMEMORY_API_KEY) {
  throw ...
}
```

with:

```ts
validateApiKey(options.apiKey)
```

This preserves the existing fail-fast behavior while allowing callers to provide the API key programmatically.

### Documentation

Updated the OpenAI JSDoc `@throws` descriptions to match the shared behavior:

> neither `options.apiKey` nor `process.env.SUPERMEMORY_API_KEY` is provided.

## Example

The following now works:

```ts
const client = withSupermemory(openai, {
  containerTag: "user-123",
  customId: "conversation-456",
  apiKey: await secrets.get("supermemory"),
})
```

Previously this threw:

```text
Error: SUPERMEMORY_API_KEY is not set
```

## Testing

Added:

- `src/openai/middleware.test.ts` (4 tests)

The tests mock the SDK and stub `fetch`, so no network access is required.

I also verified that the new tests reproduce the original bug.

Against `main`, the two new capability tests fail with:

```text
× accepts an API key from options when the env var is not set
  → expected [Function] to not throw an error but
    'Error: SUPERMEMORY_API_KEY is not set' was thrown
```

The two backward-compatibility tests continue to pass before and after the change:

- environment variable fallback
- throwing when neither `options.apiKey` nor `SUPERMEMORY_API_KEY` is provided

One test additionally verifies that the outgoing `/v4/profile` request contains:

```http
Authorization: Bearer <key>
```

confirming the supplied key is propagated all the way to the HTTP request rather than only accepted by the API surface.

## Checks

Per `CONTRIBUTING.md`:

- ✅ `vitest` — 80 passing, 0 failing assertions
- ✅ `check-types` — 146 errors on this branch, 146 on `main` (no new errors; existing ones are in `test/` files and `src/openai/tools.ts`)
- ✅ `build`
- ✅ `format-lint` (Biome clean)

## Breaking changes

None.

`apiKey` is optional, and `SUPERMEMORY_API_KEY` remains the fallback, so existing code continues to work unchanged.

The only behavioral change is that a programmatically supplied API key is now honored instead of causing an exception.


ishaanxgupta:
@rajarshidattapy please fix the conflicts

rajarshidattapy:
@ishaanxgupta give me 5 mins, on it.

rajarshidattapy:
@ishaanxgupta done, you can merge it now.

ishaanxgupta:
I still see some conflicts

rajarshidattapy:
@ishaanxgupta Thanks for catching that. I pulled the latest upstream/main again, resolved the remaining conflicts, and pushed the changes. It should be ready for another look.

MaheshtheDev:

Two things on top: we resolved the conflicts with current main (`deduplicateMemories` was renamed to `deduplicateMemoriesForMode`, and main already documents the `addMemory` default as `"always"`), and we dropped `middleware.test.ts`.

Thanks @rajarshidattapy.

PR #1248, opened 2026-07-12
fix: scope MCP document reads to the default project instead of every project

### LINKED ISSUE #1246: MCP: listMemories and memory-graph read across every project while memory and recall are scoped to the default project
## Problem

When no container tag is supplied—neither by the caller nor by the OAuth props—the MCP tools disagree about what they're operating on. Reads and writes go to the default project, but the listing and graph tools read across every project.

`SupermemoryClient` falls back to the default project in its constructor:

```ts
// apps/mcp/src/client.ts:148
this.containerTag = containerTag || DEFAULT_PROJECT_ID // "sm_project_default"
```

So `memory` and `recall`, which go through `client.createMemory()` / `client.search()` (both of which pass `containerTag: this.containerTag`), are always scoped.

But `listMemories` and `memory-graph` don't use that field—they pass `containerTags` to `getDocuments()` explicitly, and pass `undefined` when there's no tag:

```ts
// apps/mcp/src/server.ts:769-777 — handleListMemories
const effectiveContainerTag = containerTag || this.props?.containerTag // undefined

const result = await client.getDocuments(
  effectiveContainerTag ? [effectiveContainerTag] : undefined,
  page,
  limit,
)
```

Since `containerTags: undefined` is omitted from the request body, the API returns documents across **all projects**.

The same pattern exists in:

- `memory-graph` (`server.ts:350-358`)
- `fetch-graph-data` (`server.ts:422-432`)

The user-visible result is that `listMemories`—whose tool description tells the model to use it "to audit what is on file"—enumerates documents from projects that `recall` can never surface and that `memory` never writes to. `memory-graph` renders those foreign documents too.

An agent auditing memories before forgetting stale ones is shown a superset of what it can actually act on.

This is the same class of scoping bug that #1215 fixed in the browser extension.

## Steps to reproduce

Connect to the MCP server without a root `containerTag` on the OAuth props, and with memories stored in more than one project.

1. Call `memory` with:

```json
{
  "action": "save",
  "content": "..."
}
```

The document is written to `sm_project_default`.

2. Call `recall`.

Only `sm_project_default` memories are returned.

3. Call `listMemories` with no `containerTag`.

Documents from **all projects** are listed, including ones that `recall` cannot reach.

## Expected behavior

All four tools should agree on scope.

When no container tag is supplied, `listMemories`, `memory-graph`, and `fetch-graph-data` should read from `sm_project_default`, matching where `memory` writes and `recall` reads.

## Actual behavior

`listMemories`, `memory-graph`, and `fetch-graph-data` read across every project, while `memory` and `recall` are confined to `sm_project_default`.

## Suggested fix

Default the container tag in these handlers the same way the client constructor already does, rather than passing `undefined` through to the API.

The cleanest approach would be to centralize this logic in one place. `SupermemoryClient` already knows its own `containerTag`, so `getDocuments()` could fall back to that instead of requiring each handler to resolve it independently. That also prevents future handlers from reintroducing the same inconsistency.

Happy to open a PR.

## Environment

- Repo: `main @ 2cebe815`
- `apps/mcp` (server v4.0.0)
- Not platform-specific

rajarshidattapy:
## Summary

Fixes #1246.

The MCP tools disagreed about what they were operating on.

`memory` and `recall` go through `createMemory()` / `search()`, which pass `containerTag: this.containerTag`—and the client constructor defaults that to `sm_project_default` (`client.ts:148`).

But `listMemories`, `memory-graph`, and `fetch-graph-data` build their tags in the handler and pass `undefined` when none is supplied:

```ts
// server.ts:773-777
const result = await client.getDocuments(
  effectiveContainerTag ? [effectiveContainerTag] : undefined,
  page,
  limit,
)
```

Since `containerTags: undefined` is omitted from the request body, the API returned documents across **all projects**.

`listMemories`—whose tool description tells the model to use it "to audit what is on file"—was therefore enumerating documents that `recall` can never surface and that `memory` never writes to, and `memory-graph` was rendering them.

The fix defaults the tag at the single point all three tools route through, `SupermemoryClient.getDocuments`:

```ts
containerTags: containerTags ?? [this.containerTag],
```

I put it there rather than in the three handlers deliberately.

It closes the gap for all of them at once, prevents a fourth handler from reintroducing it, and makes the client internally consistent—`createMemory`, `search`, and `getProfile` already scope to `this.containerTag`, and now document reads do too.

Explicitly supplied tags are still honoured unchanged.

## Testing

Added:

- `apps/mcp/src/client.test.ts` (4 tests)

The tests stub `fetch`, so no network is required.

I also verified that the tests reproduce the original bug.

Against the unfixed code, three of the four tests fail with the reported defect:

```text
× falls back to the default project when no container tag is given
  → expected undefined to deeply equal [ 'sm_project_default' ]
```

That `undefined` is the omitted `containerTags` request body that caused the API to return every project.

The fourth test—explicitly supplied container tags are respected—passes both before and after the change, confirming the already-scoped path remains unchanged.

## Checks

- ✅ `vitest` (`apps/mcp/src/`) — 12 passing (4 new + the 8 existing in `format.test.ts`)
- ✅ `tsc --noEmit` — 0 errors
- ✅ `vite build`
- ✅ Biome — clean

I scoped test runs to `src/`; the `e2e/**` suite in `vitest.config.ts` requires live credentials.

## Breaking changes

None for correctly scoped callers.

Behavior does change for the previously unscoped path, which is the purpose of this fix: when no container tag is supplied, `listMemories`, `memory-graph`, and `fetch-graph-data` now return documents from `sm_project_default` instead of across every project.

Anyone relying on the previous cross-project listing was relying on results that `recall` and `memory` could not actually act upon.

### DIFF
--- a/apps/mcp/src/client.test.ts
+++ b/apps/mcp/src/client.test.ts
@@ -1,0 +1,71 @@
+import { afterEach, describe, expect, it, vi } from "vitest"
+import { SupermemoryClient } from "./client"
+
+const TOKEN = "sm_test_token"
+const API_URL = "https://api.supermemory.ai"
+
+/** Stubs the documents endpoint and records the request body. */
+function mockDocumentsFetch() {
+	const fetchMock = vi.fn().mockResolvedValue({
+		ok: true,
+		json: async () => ({
+			documents: [],
+			pagination: { currentPage: 1, limit: 10, totalItems: 0, totalPages: 0 },
+		}),
+	})
+	vi.stubGlobal("fetch", fetchMock)
+	return fetchMock
+}
+
+function requestBody(fetchMock: ReturnType<typeof mockDocumentsFetch>) {
+	const [, init] = fetchMock.mock.calls[0] as [string, RequestInit]
+	return JSON.parse(init.body as string) as { containerTags?: string[] }
+}
+
+afterEach(() => {
+	vi.unstubAllGlobals()
+})
+
+describe("SupermemoryClient.getDocuments container scoping", () => {
+	// listMemories, memory-graph and fetch-graph-data all pass `undefined` when no
+	// container tag is supplied. Omitting containerTags reads across every project,
+	// while createMemory/search stay scoped to the default project — so listings
+	// would surface documents that recall can never return.
+	it("falls back to the default project when no container tag is given", async () => {
+		const fetchMock = mockDocumentsFetch()
+		const client = new SupermemoryClient(TOKEN, undefined, API_URL)
+
+		await client.getDocuments(undefined, 1, 10)
+
+		expect(requestBody(fetchMock).containerTags).toEqual(["sm_project_default"])
+	})
+
+	it("falls back to the client's container tag when no container tag is given", async () => {
+		const fetchMock = mockDocumentsFetch()
+		const client = new SupermemoryClient(TOKEN, "sm_project_work", API_URL)
+
+		await client.getDocuments(undefined, 1, 10)
+
+		expect(requestBody(fetchMock).containerTags).toEqual(["sm_project_work"])
+	})
+
+	it("uses explicitly supplied container tags", async () => {
+		const fetchMock = mockDocumentsFetch()
+		const client = new SupermemoryClient(TOKEN, "sm_project_work", API_URL)
+
+		await client.getDocuments(["sm_project_other"], 1, 10)
+
+		expect(requestBody(fetchMock).containerTags).toEqual(["sm_project_other"])
+	})
+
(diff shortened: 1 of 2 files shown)


ishaanxgupta:
Hi @rajarshidattapy,
Thanks for investigating this. The current behavior is intentional: listMemories, memory-graph, and fetch-graph-data are global when containerTag is omitted. Supplying a containerTag scopes them to that project.
memory and recall have different defaults because they are project-specific operations, while listing and graph tools are intended to provide an account-wide overview.

Issue #1241, opened 2026-07-12
docs(openai): addMemory default is documented as "never" but implementation defaults to "always"


rajarshidattapy:
### Summary

The JSDoc for `addMemory` in the OpenAI middleware documents the default as `"never"`, but the implementation actually defaults to `"always"`.

This creates a mismatch between the documented behavior and the runtime behavior, especially around conversation persistence.

### Affected code

Documentation:

```
packages/tools/src/openai/middleware.ts:402
```

Implementation:

```ts
const addMemory = options?.addMemory ?? "always";
```

```
packages/tools/src/openai/middleware.ts:431
```

### Current behavior

The documentation indicates:

> default: `"never"`

However, the implementation behaves as:

```ts
options?.addMemory ?? "always"
```

meaning conversations are persisted by default unless the caller explicitly disables it.

### Why this matters

This is more than a documentation typo because it affects user expectations around persistence.

Someone reading the API docs would reasonably assume conversations are **not** stored unless they opt in, while the implementation does the opposite.

### Suggested fix

Update the JSDoc to reflect the actual default (`"always"`), or change the implementation if `"never"` was the intended behavior.

Given the current implementation, this is likely just a documentation fix.

PR #1250, opened 2026-07-12
docs(tools): correct addMemory default in OpenAI middleware JSDoc (#1241)

rajarshidattapy:
## Summary

The JSDoc for `createOpenAIMiddleware` documented `options.addMemory` as defaulting to `"never"`, but the implementation defaults to `"always"`. Updated the doc to match.

```diff
- * @param options.addMemory - Automatic memory storage mode: "always" or "never" (default: "never")
+ * @param options.addMemory - Automatic memory storage mode: "always" or "never" (default: "always")

Why docs and not the implementation

"always" is the intended default, not a bug. Every other integration in packages/tools agrees:

┌────────────────────────────┬──────────┬───────────────────────┐
│        Integration         │ Default  │     Documented as     │
├────────────────────────────┼──────────┼───────────────────────┤
│ openai/index.ts:85         │ "always" │ "always"              │
├────────────────────────────┼──────────┼───────────────────────┤
│ openai/middleware.ts:431   │ "always" │ "never" ← the outlier │
├────────────────────────────┼──────────┼───────────────────────┤
│ vercel/index.ts:141        │ "always" │ "always"              │
├────────────────────────────┼──────────┼───────────────────────┤
│ vercel/middleware.ts:257   │ "always" │ —                     │
├────────────────────────────┼──────────┼───────────────────────┤
│ voltagent/middleware.ts:81 │ "always" │ "always"              │
├────────────────────────────┼──────────┼───────────────────────┤
│ mastra/processor.ts:72     │ "always" │ —                     │
└────────────────────────────┴──────────┴───────────────────────┘

Changing the runtime default to "never" would be a breaking change to conversation persistence across all five integrations, so this is a one-line doc fix.

Closes [#1241](https://github.com/supermemoryai/supermemory/issues/1241)

### DIFF
--- a/packages/tools/src/openai/middleware.ts
+++ b/packages/tools/src/openai/middleware.ts
@@ -399,7 +399,7 @@ const addMemoryTool = async (
  * @param options.customId - Optional conversation ID to group messages for contextual memory generation
  * @param options.verbose - Enable detailed logging of memory operations (default: false)
  * @param options.mode - Memory search mode: "profile" (all memories), "query" (search-based), or "full" (both) (default: "profile")
- * @param options.addMemory - Automatic memory storage mode: "always" or "never" (default: "never")
+ * @param options.addMemory - Automatic memory storage mode: "always" or "never" (default: "always")
  * @returns Object with `wrapClient` and `createClient` methods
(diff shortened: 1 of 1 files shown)


ishaanxgupta:
Issue solved through #1253, thanks for raising the PR

PR #1251, opened 2026-07-12
fix: deduplicate @supermemory/ai-sdk into @supermemory/tools/ai-sdk

### LINKED ISSUE #1247: @supermemory/ai-sdk has drifted from @supermemory/tools and is missing its container-scoping and coercion fixes
## Problem

`packages/ai-sdk` (published as `@supermemory/ai-sdk` v1.0.8, with its own release workflow at `.github/workflows/publish-ai-sdk.yml`) reimplements what `packages/tools/src/ai-sdk.ts` already provides—and every fix that has landed in the latter has missed it.

The two are now materially different packages presenting the same API.

Three concrete divergences, each one a bug that was already fixed once:

### 1. No conflicting-config guard

`getContainerTags()` (`packages/tools/src/tools-shared.ts:64`) throws when both `projectId` and `containerTags` are passed—that was #1214.

The `ai-sdk` package silently ignores `containerTags` instead:

```ts
// packages/ai-sdk/src/tools.ts:36-38
const containerTags = config?.projectId
  ? [`sm_project_${config?.projectId}`]
  : config?.containerTags // silently dropped when projectId is also set
```

### 2. No numeric coercion

#1202 changed `limit`/`offset` to `z.coerce.number()` in `packages/tools/src/ai-sdk.ts` because models routinely emit `"5"` as a string.

`packages/ai-sdk` declares a raw JSON Schema:

```json
"type": "number"
```

(`tools.ts:57`)

which rejects exactly the input #1202 was fixed to accept.

### 3. Divergent default scope (the dangerous one)

With no config at all, `packages/tools` defaults to:

```ts
["sm_project_default"]
```

(`tools-shared.ts:72`)

while `packages/ai-sdk` leaves `containerTags` as `undefined`.

An undefined container tag isn't "the default project"; it's **no scoping**.

The same call that would read and write one project via `@supermemory/tools` searches and writes across every project via `@supermemory/ai-sdk`.

A user following the docs has no way to know which of the two packages they're getting these semantics from.

## Steps to reproduce

```ts
import { supermemoryTools } from "@supermemory/ai-sdk"

// 1. Conflicting config is accepted instead of rejected
// containerTags is silently ignored
supermemoryTools(apiKey, {
  projectId: "abc",
  containerTags: ["tag-a"],
})

// @supermemory/tools throws
// "either projectId or containerTags"
// this does not.
```

```ts
// 2. No config -> unscoped
supermemoryTools(apiKey)

// containerTags === undefined
// -> searches/writes across ALL projects

// @supermemory/tools would scope to
// ["sm_project_default"]
```

For (2), a model calling `searchMemories` with:

```ts
limit: "5"
```

is also rejected by the JSON Schema here, whereas `@supermemory/tools` correctly coerces it.

## Expected behavior

One implementation of the AI SDK tools, with one set of container-scoping and coercion semantics.

## Actual behavior

Two published packages with silently different behavior; the older one keeps missing fixes.

## Suggested fix

Two possible approaches:

- Deprecate `@supermemory/ai-sdk` and re-export from `@supermemory/tools/ai-sdk`, leaving a single implementation. This is the cleanest long-term solution, though it is a package-level deprecation and would be a breaking change for users relying on the current unscoped default.
- Port the three fixes into `packages/ai-sdk` and keep both packages. This is quicker, but the implementations will likely drift again—the same thing has already happened multiple times.

Note that fixing (3) changes behavior for existing `@supermemory/ai-sdk` users whose calls are currently unscoped, so it likely warrants a major version regardless of which approach is chosen.

Happy to implement once the direction is settled.

## Environment

- Repo: `main @ 2cebe815`
- `@supermemory/ai-sdk` 1.0.8 vs `@supermemory/tools` 2.1.0
- Not platform-specific

rajarshidattapy:
## What & why

Closes #1247.

`packages/ai-sdk` reimplemented what `packages/tools/src/ai-sdk.ts` already provides, and every fix landed in the latter missed it. Rather than porting the three fixes and waiting for the next drift, this makes `@supermemory/ai-sdk` a thin re-export of `@supermemory/tools/ai-sdk` — one implementation, one set of semantics.

The three divergences from the issue are resolved by inheritance rather than by patch:

| # | Divergence | Now |
| --- | --- | --- |
| 1 | `containerTags` silently dropped when `projectId` is also set | `getContainerTags()` throws (#1214) |
| 2 | Raw JSON Schema `"type": "number"` rejects `limit: "5"` | `z.coerce.number()` coerces it (#1202) |
| 3 | No config → `containerTags === undefined` → **unscoped** reads/writes | Defaults to `["sm_project_default"]` |

## Changes

- `packages/ai-sdk/src/tools.ts` and `tools.test.ts` deleted; `src/index.ts` is now `export * from "@supermemory/tools/ai-sdk"` with a `@deprecated` JSDoc. Coverage for this code already lives in `packages/tools` (`tools-shared.ts` / `tools.test.ts`).
- `packages/ai-sdk/package.json`: `1.0.8` → `2.0.0`; dependencies reduced to `@supermemory/tools` plus `@ai-sdk/provider` (satisfies the tools peer dep).
- `turbo.json`: `check-types` now `dependsOn: ["^build"]`. Required — `tsc` resolves `@supermemory/tools/ai-sdk` through its exports map into `dist/`, so the dependency must be built first, and CI runs `check-types --filter='@supermemory/ai-sdk'`.
- `.github/workflows/publish-ai-sdk.yml`: build step now runs `turbo run build --filter='@supermemory/ai-sdk'` from the repo root, so `@supermemory/tools` is built before the re-export's `.d.ts` is generated.
- README: deprecation banner pointing at `@supermemory/tools/ai-sdk` and listing the behavior changes.

## Breaking changes

Major bump to `2.0.0`, as the issue anticipated. Existing `@supermemory/ai-sdk` users are affected in three ways:

- Passing both `projectId` and `containerTags` now **throws** instead of ignoring `containerTags`.
- **Calls with no config are now scoped to `["sm_project_default"]`.** Previously they searched and wrote across every project. Anyone depending on that must now pass `containerTags` explicitly. This is the fix, but it will change what existing calls read and write.
- `supermemoryTools()` now returns the full `@supermemory/tools` set (`getProfile`, `documentList`, `documentAdd`, `documentDelete`, `memoryForget`) alongside `searchMemories` and `addMemory`. Additive, but it widens what a model can call if you spread the whole object into `tools`.

Follow-up for a maintainer with publish rights: run `npm deprecate @supermemory/ai-sdk` — that's a registry action, not something this PR can do.

## Verification

`turbo run build check-types --filter='@supermemory/ai-sdk'` passes; `dist/index.js` is a single re-export line. Biome clean on the changed sources.

One thing worth mentioning in the PR thread if you want to be a good citizen: building packages/tools on Windows makes tsdown's exports: true rewrite its own package.json with backslash paths ("./openai\\index"). Pre-existing, unrelated to #1247 — I reverted the file, but it's a real trap for Windows contributors.


rajarshidattapy:
@ishaanxgupta can you check this?

MaheshtheDev:
Closing this one. Four PRs are open for the same consolidation, so we are keeping #1433 as the single thread since it also refreshes the tool surface.

Worth flagging for whoever picks this up: `packages/ai-sdk/src/tools.ts` now holds a search-limit clamp added in #1593, and `packages/tools/src/ai-sdk.ts` has no equivalent. Collapsing the package into a re-export today would silently drop that bound, so the clamp needs porting across first.

PR #1252, opened 2026-07-12
fix(validation): remove dead code and no-op transforms in search schemas

### LINKED ISSUE #1191: fix(ai-sdk): use z.coerce.number() for limit/offset params to handle string inputs from LLMs

### LINKED ISSUE #1205: Dead code in search threshold schemas: unreachable `v === undefined` checks and no-op `.transform(Number)`

Paramveersingh-S:
## Description
This PR addresses dead code and redundant transformations within the Zod validation schemas in `packages/validation/api.ts`, resolving two separate but related schema issues. 

**What changed and why:**
1. **Replaced `.transform(Number)` with `z.coerce.number()`**: 
   The `limit` and `page` parameters in `ListMemoriesQuerySchema` were previously using `.string().regex(/^\d+$/).or(z.number()).transform(Number)`. This manual transformation chain was replaced with the native and safer `z.coerce.number()` to natively handle string inputs from LLMs seamlessly.
2. **Removed unreachable `v === undefined` guards**: 
   In `SearchRequestSchema` and `Searchv4RequestSchema`, the `limit` schemas included an unreachable `.refine((v) => v === undefined || (v > 0 && v <= 100))` check. Because `.default(10)` was attached, `v` would never be undefined post-parsing. This dead code was removed and replaced entirely with Zod's native `.max(100)` validator, keeping the custom error message intact.

## Verification
- Statically verified that the validation boundaries produce the exact same types.
- Ran `bun test` in `packages/validation`.
- `api.test.ts` passes flawlessly (8/8 tests passed in ~16ms), explicitly verifying that `searchSchemas` no longer contain the dead `v === undefined` range guards or `.transform(Number)`.

Fixes #1205
Fixes #1191

### DIFF
--- a/packages/validation/api.ts
+++ b/packages/validation/api.ts
@@ -271,14 +271,10 @@ export const ListMemoriesQuerySchema = z
 				}),
 			}),
 		limit: z
-			.string()
-			.regex(/^\d+$/)
-			.or(z.number())
-			.transform(Number)
-			.refine((value) => value <= 1100, {
-				message: "Limit cannot be greater than 1100",
-			})
-			.default("10")
+			.coerce
+			.number()
+			.max(1100, { message: "Limit cannot be greater than 1100" })
+			.default(10)
 			.openapi({
 				description: "Number of items per page",
 				example: "10",
@@ -288,11 +284,9 @@ export const ListMemoriesQuerySchema = z
 			.default("desc")
 			.openapi({ description: "Sort order", example: "desc" }),
 		page: z
-			.string()
-			.regex(/^\d+$/)
-			.or(z.number())
-			.transform(Number)
-			.default("1")
+			.coerce
+			.number()
+			.default(1)
 			.openapi({ description: "Page number to fetch", example: "1" }),
 		sort: z
 			.enum(["createdAt", "updatedAt"])
@@ -424,15 +418,9 @@ export const SearchRequestSchema = z.object({
 		.number()
 		.int()
 		.positive()
+		.max(100, { message: "limit must be between 1 and 100" })
 		.optional()
 		.default(10)
-		.refine((v) => v === undefined || (v > 0 && v <= 100), {
-			message: "limit must be between 1 and 100",
-			params: {
-				max: 100,
-				min: 1,
-			},
-		})
 		.openapi({
 			description: "Maximum number of results to return",
 			example: 10,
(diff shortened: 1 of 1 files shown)


ishaanxgupta:
Duplicate, issue already closed!

PR #1256, opened 2026-07-12
fix(tools): deduplicate @supermemory/ai-sdk into @supermemory/tools/a…

### LINKED ISSUE #1247: @supermemory/ai-sdk has drifted from @supermemory/tools and is missing its container-scoping and coercion fixes

Paramveersingh-S:
## Summary
`packages/ai-sdk` had aggressively drifted from `packages/tools/src/ai-sdk.ts`, missing several major bug fixes related to default container tagging and Zod coercion (e.g., #1202, #1214).

**Changes:**
- Deleted the duplicated and outdated `tools.ts` logic inside `packages/ai-sdk`.
- Swapped `packages/ai-sdk/src/index.ts` to be a pure passthrough re-export: `export * from "@supermemory/tools/ai-sdk"`.
- Cleaned up the local dependencies in `package.json` to only rely on `@supermemory/tools` and `@ai-sdk/provider`, bumping the version to `v2.0.0` to reflect the inherited behavior fixes.

There is now only one implementation dictating scoping semantics.

Fixes #1247


MaheshtheDev:
Closing this one. Four PRs are open for the same consolidation, so we are keeping #1433 as the single thread since it also refreshes the tool surface.

Worth flagging for whoever picks this up: `packages/ai-sdk/src/tools.ts` now holds a search-limit clamp added in #1593, and `packages/tools/src/ai-sdk.ts` has no equivalent. Collapsing the package into a re-export today would silently drop that bound, so the clamp needs porting across first.

PR #1260, opened 2026-07-12
fix(web): declare entityContext in bulkLinkMutation so bulk link import works

### LINKED ISSUE #1259: Bulk multi-URL paste always fails: bulkLinkMutation references an undeclared entityContext
**Environment:** apps/web on current `main`, any browser.

Pasting multiple URLs into the add-memory flow (the bulk multi-URL paste added in #1109) always fails with the generic "Failed to add links" toast. No request ever reaches the API.

**Why it happens**

`bulkLinkMutation` in `apps/web/hooks/use-document-mutations.ts` references `entityContext` when building each batched document body:

```ts
documents: chunk.map((url) => ({
    content: url,
    containerTags: [project],
    entityContext,          // <- no declaration anywhere in this function
    metadata: { sm_source: "consumer" },
})),
```

Unlike its `noteMutation` / `linkMutation` / `fileMutation` siblings, the function never declares `const entityContext = await resolveEntityContext(project)`, so the identifier is unbound and the `mutationFn` throws `ReferenceError: entityContext is not defined` before the `$fetch` call. The throw is caught by the mutation's `onError`, which rolls back the optimistic entries and shows the generic failure toast — making it look like an API error.

`tsc` reports it as TS2304 (`Cannot find name 'entityContext'`), but `apps/web` builds with `ignoreBuildErrors`, so it slipped through.

**Steps to reproduce**

1. Open the add-memory modal in the web app
2. Paste two or more URLs (newline or space separated) and submit

Expected: the URLs are ingested as separate memories via `POST /documents/batch`.
Actual: immediate "Failed to add links" toast; the network tab shows no batch request was made.

Fix is a two-liner (declare it the same way the sibling mutations do and forward it with the conditional spread) — PR incoming.

abhay-codes07:
Fixes #1259

## What

The bulk multi-URL paste flow (added in #1109) has been throwing before sending any request: `bulkLinkMutation` uses `entityContext` when building each batched document body, but never declares it. Every bulk submit hit `ReferenceError: entityContext is not defined`, which the mutation's `onError` surfaced as the generic "Failed to add links" toast after rolling back the optimistic entries — so it looked like an API failure, not a client crash.

`tsc` does flag this as TS2304, but `apps/web` builds with `ignoreBuildErrors`, which is how it survived.

## Fix

Resolve the entity context once at the top of the `mutationFn`, exactly like the `noteMutation`/`linkMutation`/`fileMutation` siblings, and forward it with the same conditional spread so the key is omitted when the space already has its own entity context configured:

```ts
const entityContext = await resolveEntityContext(project)
...
...(entityContext !== undefined ? { entityContext } : {}),
```

## Testing

- `bun test` in `apps/web` — 37 pass
- `bun run build` in `apps/web` — clean
- `biome check` on the touched file — clean

cc @MaheshtheDev

### DIFF
--- a/apps/web/hooks/use-document-mutations.ts
+++ b/apps/web/hooks/use-document-mutations.ts
@@ -390,6 +390,7 @@ export function useDocumentMutations({
 			urls: string[]
 			project: string
 		}): Promise<{ success: number; failed: number }> => {
+			const entityContext = await resolveEntityContext(project)
 			let success = 0
 			let failed = 0
 
@@ -400,7 +401,7 @@ export function useDocumentMutations({
 						documents: chunk.map((url) => ({
 							content: url,
 							containerTags: [project],
-							entityContext,
+							...(entityContext !== undefined ? { entityContext } : {}),
 							metadata: { sm_source: "consumer" },
 						})),
 					},


ishaanxgupta (approved):

ishaanxgupta:
Thank you for the contribution @abhay-codes07

PR #1261, opened 2026-07-12
fix(extension): repair Included Memories popup data handling and lifecycle

### LINKED ISSUE #1257: Extension: Included Memories popup fragments comma-containing memories, leaks global click listeners, and drops the last memory on removal
**Environment:** browser extension on current `main` (wxt, chrome-mv3). The popup block is duplicated for ChatGPT, Claude and T3, so all three surfaces are affected.

While reading through the related-memories flow in the content scripts, I found a cluster of related problems in the popup pipeline shared by `chatgpt.ts`, `claude.ts` and `t3.ts` — each one traceable directly in the code:

**1. Memories containing commas get fragmented into fake rows**

`getRelatedMemories` in `background.ts` returns `data` as a `string[]`. The content scripts assign that array straight onto a dataset attribute (`iconElement.dataset.memoriesData = response.data`), which coerces it to a single comma-joined string. The popup then re-splits it with `.split(/[,\n]/)`.

Steps to reproduce:
1. Save a memory whose content contains a comma, e.g. `Prefers TypeScript, strict mode, and Biome`
2. Type a related prompt on chatgpt.com so auto-search picks it up
3. Click "Included Memories"

Expected: one row for the memory.
Actual: three rows ("Prefers TypeScript" / "strict mode" / "and Biome"). Since the remove buttons operate by row index into that mis-split array, removing a row deletes the wrong fragment and rewrites `dataset.supermemories` into text that no longer matches what the popup shows.

**2. Removing one of two memories throws away the last one too**

The cleanup guard after a removal is `if (currentMemories.length <= 1)`, which wipes `dataset.supermemories`, clears the icon, and destroys the popup. So with two memories included, removing one also silently discards the one you wanted to keep. Should be `=== 0`.

**3. Every search leaks a popup node and a global click listener**

Each successful fetch appends a new popup `div` to `document.body` and registers a new `document.addEventListener("click", ...)` closing over it. The listener is never removed and the popup only goes away after a 5-minute timeout. With auto-search firing on every debounced input change, editing a prompt for a few minutes accumulates dozens of orphaned global click handlers and stacked hidden popups.

**4. Prompt injection text degrades after removals**

After a removal the remaining memories are re-joined with `" ,"` and the numbering baked into the strings by the background handler goes stale, so the injected "Supermemories of user" block ends up as `1. foo , 3. bar`.

I have a fix ready that stores the array as JSON on the dataset (so content survives the round-trip), rebuilds the prompt block from that single source of truth, fixes the `<= 1` guard, and disposes the previous popup/listener before mounting a new one — PR incoming.

abhay-codes07:
Fixes #1257

## What

The "Included Memories" popup pipeline on ChatGPT/Claude/T3 had four related defects, all rooted in the same ~180-line block duplicated across the three content scripts:

1. **Comma fragmentation.** The background script returns memories as a `string[]`; assigning that to `dataset.memoriesData` coerces it into one comma-joined string, and the popup re-split it with `.split(/[,\n]/)`. Any memory containing a comma showed up as several fake rows, and the index-based remove buttons spliced the wrong entries while rewriting a desynced prompt injection.
2. **Last-memory wipe.** The removal cleanup used `length <= 1`, so removing one of two memories also silently discarded the remaining one.
3. **Listener/DOM leak.** Every successful fetch appended a fresh popup to `<body>` and registered a new `document`-level click listener that was never removed. With auto-search firing per debounced input change, a few minutes of prompt editing accumulates dozens of orphaned global click handlers and stacked popups.
4. **Degraded prompt text.** Removals re-joined the memories with `" ,"` and the numbering baked in by the background handler went stale, leaving the injected block as `1. foo , 3. bar`.

## Fix

Since the block was byte-identical in `chatgpt.ts`, `claude.ts`, and `t3.ts` (apart from the prompt-element lookup), the popup now lives in a shared `utils/included-memories.ts`:

- memories are stored as **JSON** on the dataset, so content survives the round-trip regardless of what characters it contains
- the injected `dataset.supermemories` block is always rebuilt from that single source of truth, with numbering applied at build time so it stays correct after removals
- rows re-render from the array instead of juggling `data-memory-index` attributes
- removing entries keeps the last one until none remain (`=== 0`), and only then clears the injection and restores the icon
- each popup disposes its predecessor (node, document listener, auto-dismiss timer) before mounting, so repeated auto-searches no longer leak

The background handler now returns raw memory strings and skips empty results; each content script passes its own site-specific prompt-element lookup. Net −206 lines.

## Testing

- `bun run compile` (tsc --noEmit) — clean
- `bun run build` (wxt, chrome-mv3) — clean
- `biome check` on all touched files — clean

The extension package has no test harness, so the helper was written as pure functions (`serializeMemories`/`parseMemories`/`buildPromptInjection`) plus a self-contained popup factory to keep the behaviour easy to reason about and to unit-test later if a harness lands. Happy to add screenshots/screen recording of the popup flow if that helps review.

cc @MaheshtheDev

### DIFF
--- a/apps/browser-extension/entrypoints/background.ts
+++ b/apps/browser-extension/entrypoints/background.ts
@@ -162,9 +162,12 @@ export default defineBackground(() => {
 			const response = responseData as {
 				results?: Array<{ memory?: string }>
 			}
+			// Return raw memory strings; consumers handle numbering/formatting.
 			const memories: string[] = []
-			response.results?.forEach((result, index) => {
-				memories.push(`${index + 1}. ${result.memory} \n`)
+			response.results?.forEach((result) => {
+				if (result.memory) {
+					memories.push(result.memory)
+				}
 			})
 			console.log("Memories:", memories)
 			await trackEvent(eventSource)
--- a/apps/browser-extension/entrypoints/content/chatgpt.ts
+++ b/apps/browser-extension/entrypoints/content/chatgpt.ts
@@ -10,6 +10,11 @@ import {
 	autoCapturePromptsEnabled,
 } from "../../utils/storage"
 import {
+	buildPromptInjection,
+	createIncludedMemoriesPopup,
+	serializeMemories,
+} from "../../utils/included-memories"
+import {
 	createChatGPTInputBarElement,
 	DOMUtils,
 } from "../../utils/ui-components"
@@ -156,16 +161,21 @@ async function getRelatedMemoriesForChatGPT(actionSource: string) {
 			timeoutPromise,
 		])
 
-		if (response?.success && response?.data) {
+		if (
+			response?.success &&
+			Array.isArray(response.data) &&
+			response.data.length > 0
+		) {
+			const memories = response.data as string[]
 			const promptElement = document.getElementById("prompt-textarea")
 			if (promptElement) {
-				promptElement.dataset.supermemories = `\n\nSupermemories of user (only for the reference): ${response.data}`
+				promptElement.dataset.supermemories = buildPromptInjection(memories)
 				console.log(
 					"Prompt element dataset:",
 					promptElement.dataset.supermemories,
 				)
 
-				iconElement.dataset.memoriesData = response.data
+				iconElement.dataset.memoriesData = serializeMemories(memories)
 
 				updateChatGPTIconFeedback("Included Memories", iconElement)
 			} else {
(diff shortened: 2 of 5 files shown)


ishaanxgupta:
Hi @abhay-codes07 can you please fix the merge conflicts?

abhay-codes07:
Looking at the rebase, the conflicts are not mechanical. The Included Memories popup was reworked on `main` since this PR opened (the new `entrypoints/content/memory-suggestion.ts` with `showMemorySuggestion` / `showMarkerPopover` / `setMemoryMarkerStatus`), which covers most of the lifecycle restructuring this PR was doing. So the shared-util + `createIncludedMemoriesPopup` approach here is largely superseded, and a straight conflict resolution would just fight the new module. Happy to drop that part.

One thing did survive the refactor though, and it is the original data-handling bug from #1257. In the new code the memories are stored as a comma-joined string and split back apart on `[,\n]`:

- store: `iconElement.dataset.memoriesData = String(response.data)` (chatgpt.ts:215, claude.ts:462, t3.ts:236)
- read: `memories.split(/[,\n]/)` in `showMarkerPopover` (memory-suggestion.ts:308-311), and the same split in t3's popup list and delete handler (t3.ts:332, 424)

Because memory text itself can contain commas and newlines, a single memory gets shredded. For example "I live in Austin, Texas" renders as two separate rows ("I live in Austin" and "Texas"), and t3's per-item delete then splices the wrong index and rewrites the injected `supermemories` prompt from the mangled list.

The focused fix on top of the new module is to make the round-trip delimiter-safe: serialize `response.data` as JSON at the store sites and parse it back in `showMarkerPopover` (with a fallback to the old split so any already-stored value still renders). t3 needs the same parse in its list/delete path plus keeping its injected-prompt string derived from the parsed array rather than the raw join.

I would rather rebuild it that way than force this PR's old structure through. I can either repoint this PR to that focused change, or close it and open a fresh one against `memory-suggestion.ts`, whichever you prefer. Let me know and I will push it.
@ishaanxgupta

ishaanxgupta:
The popup fixes are useful, but this PR changes the shared GET_RELATED_MEMORIES response format without updating every consumer. When tested with the current main branch, this caused auto-recall regressions. This should preserve the existing response contract or update all ChatGPT, Claude, Gemini, and T3 consumers together with integration tests. Closing this version for now.

abhay-codes07:
Understood, thanks for the detail on the auto-recall regression. Opened #1339 as the contract-preserving version: it leaves the GET_RELATED_MEMORIES response and the injected `dataset.supermemories` prompt untouched, and fixes only the comma/newline fragmentation inside the popup's own `dataset.memoriesData` by storing it as JSON instead of a joined string. All four consumers (ChatGPT, Claude, Gemini, T3) are updated together and there are unit tests for the round-trip.

PR #1262, opened 2026-07-12
fix(memory-graph): allow selecting nodes on touch devices

### LINKED ISSUE #1258: Memory graph: nodes cannot be selected on touch devices
**Environment:** any touch device / touch-event browser, `@supermemory/memory-graph` on current `main`.

On touch devices it is impossible to select a node in the memory graph: there is no code path that can produce a selection from a touch. Tapping a document or memory node does nothing — no selection, no popover, no way to open a document from the graph. Pan and pinch-zoom work.

**Why it happens**

Node selection is driven by the canvas `click` handler in `packages/memory-graph/src/canvas/input-handler.ts`. On touch browsers that click never arrives: `onTouchStart` calls `e.preventDefault()` (needed to stop the page from scrolling/zooming), and per the spec that also suppresses the browser-synthesized mouse events, including `click`. There is no touch-based tap detection to compensate — a single-finger `touchstart` goes straight into `isPanning = true`.

**Steps to reproduce**

1. Open the memory graph on a phone (or DevTools device emulation with touch events)
2. Tap any node

Expected: the node gets selected and its popover opens (same as a mouse click).
Actual: nothing happens. The only interactions available on touch are pan and pinch.

Happy to submit a fix — I have tap detection working locally (track the single-finger start point, cancel the tap once the finger moves past a small slop threshold or a second finger lands, and hit-test through the viewport transform on release).

abhay-codes07:
Fixes #1258

## What

Selecting a node in the memory graph was impossible on touch devices. Selection rides on the canvas `click` event, but `onTouchStart` calls `preventDefault()` (required to keep the page from scrolling/zooming), which also suppresses the browser-synthesized mouse events — so the click never fires. A single-finger touch went straight into panning, leaving phones and tablets with pan/pinch only: no selection, no popover, no way to open a document from the graph.

## Fix

Detect taps from the raw touch lifecycle in `InputHandler`:

- remember where a single-finger touch starts
- cancel the tap once the finger travels more than a 10px slop threshold or a second finger lands (pinch)
- on release, hit-test the start point through the current viewport transform and forward the result to `onNodeClick`

Tapping a node selects it (popover opens); tapping empty space clears the selection — matching mouse behaviour. Pan and pinch are untouched.

## Testing

- New unit tests drive the handler through a stub canvas and real `ViewportState`/`SpatialIndex` instances: tap on node, tap on empty space, jitter within the slop threshold, pan, pinch, and a tap through a zoomed/panned viewport (`src/__tests__/input-handler-touch.test.ts`)
- `bun run test` in `packages/memory-graph` — 194 pass

cc @MaheshtheDev

### DIFF
--- a/packages/memory-graph/src/__tests__/input-handler-touch.test.ts
+++ b/packages/memory-graph/src/__tests__/input-handler-touch.test.ts
@@ -1,0 +1,160 @@
+import { beforeEach, describe, expect, it } from "vitest"
+import { SpatialIndex } from "../canvas/hit-test"
+import { InputHandler } from "../canvas/input-handler"
+import { ViewportState } from "../canvas/viewport"
+import type { GraphNode } from "../types"
+
+/**
+ * InputHandler tap-to-select on touch devices.
+ *
+ * onTouchStart calls preventDefault() (required to stop scroll/zoom of the
+ * page), which also suppresses the browser's synthesized click event — so
+ * node selection must be detected from the raw touch lifecycle. These tests
+ * drive the handler through a stub canvas that records the listeners it
+ * registers.
+ */
+
+type Listener = (e: Event) => void
+
+function makeStubCanvas() {
+	const listeners = new Map<string, Listener>()
+	const canvas = {
+		addEventListener: (name: string, fn: Listener) => {
+			listeners.set(name, fn)
+		},
+		removeEventListener: (name: string) => {
+			listeners.delete(name)
+		},
+		getBoundingClientRect: () => ({
+			left: 0,
+			top: 0,
+			right: 800,
+			bottom: 600,
+			width: 800,
+			height: 600,
+			x: 0,
+			y: 0,
+		}),
+		style: {} as CSSStyleDeclaration,
+	}
+	return { canvas: canvas as unknown as HTMLCanvasElement, listeners }
+}
+
+function touch(clientX: number, clientY: number) {
+	return { clientX, clientY }
+}
+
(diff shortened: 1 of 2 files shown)


ishaanxgupta:
Hi @abhay-codes07,
During a small movement below `TAP_MOVE_THRESHOLD`, `onTouchMove` still pans the viewport. Then `onTouchEnd ` hit-tests `touchStartX/touchStartY` using the already-updated viewport transform. This means the original screen point no longer necessarily maps to the node that was tapped, especially when the graph is zoomed out. A normal few-pixel finger jitter can therefore still cause the tap to miss.

Please either:

- store the world coordinate/target node at touchstart and use it on release, or
- hit-test using the latest touch position (lastMouseX/lastMouseY) after the viewport has moved.

abhay-codes07:
Good catch, you are right. The sub-threshold pan in `onTouchMove` moves the viewport, and `onTouchEnd` was re-projecting `touchStartX/touchStartY` through that already-moved transform, so at low zoom a few-pixel jitter shifts the hit point off the node.

Fixed by going with your first option: I capture the world point under the finger at touchstart (before any panning) and hit-test against that on release, instead of re-projecting the start screen point. The stored world coordinate is stable regardless of how the viewport pans during the tap.

Added a regression test that reproduces exactly this: at zoom 0.25 an 8px jitter (below the 10px threshold) pans the viewport, and the old re-projection landed 32 world units off the node (past the ~20 unit hit radius) and missed. It now resolves the node correctly. Pushed in ad70ef8d.

ishaanxgupta (approved):

PR #1263, opened 2026-07-12
fix(memory-graph): keep node selection when pagination changes the node count

abhay-codes07:
## What

Pin a node's popover in the memory graph, zoom out far enough for the load-more trigger to fetch the next page, and the popover vanishes mid-read.

The slideshow effect in `memory-graph.tsx` depends on `nodes.length`, and its inactive branch unconditionally cleared the selected node:

```ts
useEffect(() => {
    if (!isSlideshowActive || nodes.length === 0) {
        if (!isSlideshowActive) {
            setSelectedNode(null)   // runs on every node-count change
            ...
```

With the slideshow off (the default for every consumer), any change in node count re-ran the effect and wiped the user's selection — pagination appending documents being the common trigger. It also fired a stray `coolDown()` on the simulation each time.

## Fix

Track the previous slideshow state in a ref and only clear the selection on a genuine active → inactive transition. Re-runs caused by `nodes.length` changing while the slideshow was never active are now no-ops.

## Testing

- `bun run test` in `packages/memory-graph` — 188 pass

Behaviour notes for review: the active → inactive transition still clears the selection and cools the simulation exactly as before; the only change is that re-runs where the slideshow was never active no longer touch the selection.

### DIFF
--- a/packages/memory-graph/src/components/memory-graph.tsx
+++ b/packages/memory-graph/src/components/memory-graph.tsx
@@ -609,14 +609,23 @@ export function MemoryGraph({
 	const onSlideshowNodeChangeRef = useRef(onSlideshowNodeChange)
 	onSlideshowNodeChangeRef.current = onSlideshowNodeChange
 
+	// Track the previous slideshow state so deactivation can be distinguished
+	// from unrelated re-runs (e.g. nodes.length changing when pagination
+	// appends documents). Only a true active -> inactive transition should
+	// clear the user's selection.
+	const wasSlideshowActiveRef = useRef(false)
+
 	useEffect(() => {
-		if (!isSlideshowActive || nodes.length === 0) {
-			if (!isSlideshowActive) {
+		if (!isSlideshowActive) {
+			if (wasSlideshowActiveRef.current) {
+				wasSlideshowActiveRef.current = false
 				setSelectedNode(null)
 				simulationRef.current?.coolDown()
 			}
 			return
 		}
+		wasSlideshowActiveRef.current = true
+		if (nodes.length === 0) return
 
 		let lastIdx = -1
 		let coolDownTimer: ReturnType<typeof setTimeout> | null = null


ishaanxgupta:
I don't think this issue exists in prod, I checked and could not reproduce this closing this for now, if you find this issue in prod, feel free to tag me

abhay-codes07:
@ishaanxgupta thanks for checking. Taking you up on the "tag me" offer, because I think this reproduces on the default path rather than through the slideshow.

The trigger is not the slideshow being on, it is the effect re-running while the slideshow is **off**. On current `main`, `memory-graph.tsx`:

```ts
useEffect(() => {
    if (!isSlideshowActive || nodes.length === 0) {
        if (!isSlideshowActive) {
            setSelectedNode(null)      // clears selection on *every* re-run
            simulationRef.current?.coolDown()
        }
        return
    }
    // ...
}, [isSlideshowActive, nodes.length])   // re-runs whenever nodes.length changes
```

`isSlideshowActive` defaults to `false` for every consumer, so the effect sits in that top branch. Because `nodes.length` is a dependency, the effect re-runs any time the node count changes, and each re-run unconditionally calls `setSelectedNode(null)`.

Repro without the slideshow:

1. Open the graph normally (slideshow off).
2. Click a node so its popover is pinned (`selectedNode` is set).
3. Do anything that appends nodes, the common one being the zoom-out load-more that fetches the next page. `nodes.length` goes up.
4. The effect re-runs, hits `setSelectedNode(null)`, and the pinned popover disappears mid-read. It also fires a stray `coolDown()` on the simulation each time.

It is easy to miss because it only bites when there is a pinned selection *and* the node count changes in the same session, so a quick click-test without pagination looks fine. The PR keeps the active → inactive transition behaving exactly as before and only stops the no-op re-runs from clearing the selection.

If that repro holds up on your end, would you mind reopening? Happy to rebase onto the latest `main` first.

PR #1264, opened 2026-07-12
fix(memory-graph): popover fallback now picks the side with the most space

abhay-codes07:
## What

`pickBestQuadrant` decides which side of a node the hover popover opens on. When no side fully fits the popover, it's supposed to fall back to the side with the most room — but the fit check flattened each side's measured space to `-1` before the fallback sort ran:

```ts
const fits: [Quadrant, number][] = [
    ["right", spaceRight >= popoverWidth ? spaceRight : -1],
    ...
]
...
return fits.sort((a, b) => b[1] - a[1])[0]?.[0] ?? "right"
```

Once every entry is `-1` the sort is meaningless and the fallback always answers `"right"` — even for a node hugging the right edge of a small container where the left side has ~45× more space. The clamping that follows then slides the popover over the node it's annotating.

## Fix

Keep the raw space measurement alongside the required size for each side. The preference pass (`right → left → below → above`) still returns the first side that fully fits — behaviour there is unchanged, including exact-fit boundaries — but the fallback can now genuinely rank sides by available space.

The helper is exported so it can be unit-tested directly.

## Testing

- New tests in `src/__tests__/popover-quadrant.test.ts` cover the preference order, the exact-fit boundary, and three fallback scenarios (node hugging the right edge / top / bottom) that the old code all answered with `"right"`
- `bun run test` in `packages/memory-graph` — 196 pass

### DIFF
--- a/packages/memory-graph/src/__tests__/popover-quadrant.test.ts
+++ b/packages/memory-graph/src/__tests__/popover-quadrant.test.ts
@@ -1,0 +1,90 @@
+import { describe, expect, it } from "vitest"
+import { pickBestQuadrant } from "../components/node-hover-popover"
+
+/**
+ * pickBestQuadrant(screenX, screenY, nodeRadius, containerWidth,
+ * containerHeight, popoverWidth, popoverHeight)
+ *
+ * gap = 24 in the implementation, so available space per side is:
+ *   right: containerWidth - (screenX + nodeRadius + 24)
+ *   left:  screenX - nodeRadius - 24
+ *   above: screenY - nodeRadius - 24
+ *   below: containerHeight - (screenY + nodeRadius + 24)
+ */
+describe("pickBestQuadrant", () => {
+	const RADIUS = 10
+	const POPOVER_W = 476 // CARD_W(280) + 12 + SHORTCUTS_W(160) + GAP(24)
+	const POPOVER_H = 230
+
+	it("prefers right when there is room on the right", () => {
+		expect(
+			pickBestQuadrant(100, 300, RADIUS, 1200, 600, POPOVER_W, POPOVER_H),
+		).toBe("right")
+	})
+
+	it("falls back to left when only the left side fits", () => {
+		// right space: 1200 - (1100 + 10 + 24) = 66 -> no fit
+		// left space: 1100 - 10 - 24 = 1066 -> fits
+		expect(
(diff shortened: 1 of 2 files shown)


ishaanxgupta:
Hi @abhay-codes07 I could not reproduce this issue in prod, can you share a video for this?

abhay-codes07:
It is hard to catch in prod because it only shows up when the popover fits on no side at all (small viewport, or a node crowded near an edge), and even then it just picks a worse side rather than breaking outright. Since `pickBestQuadrant` is a pure function now, here is a deterministic repro with concrete numbers rather than a video.

On `main` the "does it fit" check overwrites the real available space with `-1` before the fallback sort runs:

```ts
const fits: [Quadrant, number][] = [
  ["right", spaceRight >= popoverWidth ? spaceRight : -1],
  ["left",  spaceLeft  >= popoverWidth ? spaceLeft  : -1],
  ["above", spaceAbove >= popoverHeight ? spaceAbove : -1],
  ["below", spaceBelow >= popoverHeight ? spaceBelow : -1],
]
// ...
return fits.sort((a, b) => b[1] - a[1])[0]?.[0] ?? "right"
```

When nothing fits, every entry is `-1`, so the sort has nothing to order and returns the first element, which is always `right`.

Concrete case, `pickBestQuadrant(screenX=360, screenY=200, nodeRadius=20, containerWidth=400, containerHeight=400, popoverWidth=320, popoverHeight=200)`:

- real space: right `-4`, left `316`, above `156`, below `156`
- popover needs 320 wide / 200 tall, so nothing fits fully
- `main` returns `right` (space `-4`, the single most cramped side, so the popover clamps straight over the node)
- the side with the most room is clearly `left` at `316`

With the fix, the available space is kept separate from the required space, so the fallback sorts on the true numbers and returns `left`. This exact case is in the new `popover-quadrant.test.ts`. So a node near the right edge of a narrow container pops its card to the right, on top of itself, instead of to the open space on the left.

ishaanxgupta:
Yes so like thats why I was saying that it is actually not an issue, closing this for now, feel free to ping me if you find something!

abhay-codes07:
Sounds good, thanks for taking the time to look at it. Makes sense that the downstream clamping keeps it in bounds in practice, so the fallback branch picking the more cramped side is not visible. Closing is fine by me. I will ping you if I hit a case where it actually shows up.

PR #1265, opened 2026-07-12
fix(web): reset command palette selection when the result list changes

abhay-codes07:
## What

The documents command palette keeps a `selectedIndex` into its item list, and the effect commented "Reset selection on items change" has an empty dependency array — so it only ever runs on mount:

```ts
// Reset selection on items change
useEffect(() => {
    setSelectedIndex(0)
}, [])
```

Arrow down to row 5 of the recent-documents list, then type a query that returns two results: `selectedIndex` stays 5, the highlight disappears (no row renders `data-index="5"`), and Enter silently does nothing because `items[5]` is `undefined`. The selection also never snaps back to the top result when fresh results stream in.

## Fix

Two small effects that match the comment's intent:

- snap the selection to the top whenever the query text changes
- clamp it into range whenever the item list shrinks (async search results replacing the recent docs, refetches while the palette is open)

Clamping rather than always resetting keeps the cursor position stable when the list merely refreshes without shrinking.

## Testing

- `bun test` in `apps/web` — 37 pass
- `bun run build` in `apps/web` — clean
- `biome check` on the touched file — clean (the one remaining warning about `items` in `handleKeyDown` deps predates this change)

### DIFF
--- a/apps/web/components/documents-command-palette.tsx
+++ b/apps/web/components/documents-command-palette.tsx
@@ -199,10 +199,16 @@ export function DocumentsCommandPalette({
 		for (const a of actions) items.push(a)
 	}
 
-	// Reset selection on items change
+	// biome-ignore lint/correctness/useExhaustiveDependencies: search is the trigger — snap the selection back to the top result whenever the query changes
 	useEffect(() => {
 		setSelectedIndex(0)
-	}, [])
+	}, [search])
+
+	// Keep the selection in range when the list shrinks (async results
+	// replacing recent docs, refetches, etc.)
+	useEffect(() => {
+		setSelectedIndex((i) => Math.min(i, Math.max(items.length - 1, 0)))
+	}, [items.length])
 
 	// Scroll selected into view
 	useEffect(() => {


ishaanxgupta:
Hi @abhay-codes07 I tried to replicate this bug but could not find the issue, can you share the video recording for this?

abhay-codes07:
Sure, here is a deterministic repro without needing a recording. It is easy to miss because it only bites after you have moved the selection and then the list shrinks under it.

On `main` the reset effect is keyed on an empty dependency array, so it only ever runs once on mount:

```ts
// Reset selection on items change
useEffect(() => {
  setSelectedIndex(0)
}, [])
```

`selectedIndex` is then never brought back into range when the query changes or the result list gets shorter. Steps:

1. Open the palette and type a query that returns several results.
2. Press ArrowDown a few times to highlight, say, the 5th row (`selectedIndex` is 4).
3. Refine the query so it now matches only one or two results.

At this point `selectedIndex` is still 4, which is past the end of the new list. Two visible symptoms:

- The highlight disappears, because no row has `data-index="4"` anymore (the `isSelected = index === selectedIndex` check never matches).
- Pressing Enter does nothing: `const item = items[selectedIndex]` is `items[4]`, which is `undefined`, so `if (item)` is false and the palette just sits there until you arrow back up into range.

The async path triggers it too: you type, the synchronous recent-docs list renders long, then the debounced search resolves 250ms later with fewer results and the stale index is left dangling.

The PR fixes both by keying the reset on `search` (snap to the top result when the query changes) and clamping the index to `items.length - 1` when the list shrinks. Let me know if you want me to add a small test around the clamp.

PR #1269, opened 2026-07-12
fix(extension): paginate Twitter bookmark-folder imports past the first page

### LINKED ISSUE #1267: Extension: Twitter bookmark-folder import silently caps at the first page
**Environment:** browser extension on current `main` (Twitter/X bookmarks import).

Importing a bookmark **folder** (collection) silently imports only the first API page and then reports the truncated count as a successful, complete import. Regular all-bookmarks import paginates fine.

Three things in the import path conspire (`utils/twitter-import.ts`, `utils/twitter-utils.ts`):

1. The pagination recursion is explicitly gated off for folders:
   ```ts
   if (nextCursor && tweets.length > 0 && !this.config.isFolderImport) {
   ```
   so a folder import never follows the cursor at all.

2. Even if it did, `buildBookmarkCollectionVariables(bookmarkCollectionId)` has no cursor parameter — every request would just re-fetch page 1. (`buildRequestVariables` for the regular flow does thread the cursor.)

3. The URL construction for the folder-with-cursor case appends `&variables=...` twice (the inner ternary already includes it, then the outer template appends it again), which would produce a malformed request the moment the recursion was enabled.

**Steps to reproduce:** put 100+ tweets in a bookmark folder, run the folder import from the extension → `onComplete` reports roughly one page's worth (about the first batch), remainder never imported, no error surfaced.

All three are small, related fixes — PR incoming.

abhay-codes07:
Fixes #1267

## What

Importing a Twitter bookmark **folder** silently capped at the first API page while reporting the truncated count as a complete import. Three related defects in the import path:

1. the pagination recursion was gated on `!this.config.isFolderImport`, so folder imports never followed the cursor
2. `buildBookmarkCollectionVariables` had no cursor parameter, so even with recursion every request would re-fetch page 1
3. the folder branch of the URL construction appended `&variables=...` twice when a cursor was present (the inner ternary already included it, then the outer template appended it again)

## Fix

- `buildBookmarkCollectionVariables(bookmarkCollectionId, cursor?)` threads the cursor exactly like `buildRequestVariables` does for the regular flow
- the recursion guard drops the folder exclusion; termination works the same way as the all-bookmarks flow (the last page yields a cursor but zero tweets, which stops the recursion)
- the nested URL ternaries collapse into a single `baseUrl + variables` expression, which removes the double-append path entirely

The 1s inter-page delay and 429 backoff apply to folder pages the same as regular pages.

## Testing

- `bun run compile` (tsc --noEmit) — clean
- `bun run build` (wxt, chrome-mv3) — clean
- `biome check` on both touched files — clean

I don't have a Twitter account with a 100+ tweet bookmark folder to drive the full flow end-to-end, so a maintainer with one may want to sanity-check the second page lands; the cursor extraction (`extractNextCursor`) and tweet parsing (`getAllTweets`) already handle the folder timeline shape and are unchanged.

cc @MaheshtheDev

### DIFF
--- a/apps/browser-extension/utils/twitter-import.ts
+++ b/apps/browser-extension/utils/twitter-import.ts
@@ -113,19 +113,14 @@ export class TwitterImporter {
 			const headers = createTwitterAPIHeaders(tokens)
 
 			// Build API request with pagination
-			const variables =
-				this.config.isFolderImport && this.config.bookmarkCollectionId
-					? buildBookmarkCollectionVariables(this.config.bookmarkCollectionId)
-					: buildRequestVariables(cursor)
-			const urlWithCursor = cursor
-				? `${
-						this.config.isFolderImport && this.config.bookmarkCollectionId
-							? `${BOOKMARK_COLLECTION_URL}&variables=${encodeURIComponent(JSON.stringify(variables))}`
-							: BOOKMARKS_URL
-					}&variables=${encodeURIComponent(JSON.stringify(variables))}`
-				: this.config.isFolderImport && this.config.bookmarkCollectionId
-					? `${BOOKMARK_COLLECTION_URL}&variables=${encodeURIComponent(JSON.stringify(variables))}`
-					: `${BOOKMARKS_URL}&variables=${encodeURIComponent(JSON.stringify(variables))}`
+			const collectionId = this.config.isFolderImport
+				? this.config.bookmarkCollectionId
+				: undefined
+			const variables = collectionId
+				? buildBookmarkCollectionVariables(collectionId, cursor)
+				: buildRequestVariables(cursor)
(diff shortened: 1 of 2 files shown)


ishaanxgupta:
Hi @abhay-codes07 please fix the merge conflicts

abhay-codes07:
Rebased onto latest `main`, conflicts resolved. The only overlap was the pagination guard block: `main` had since dropped the two debug `console.log` lines, and this PR removes the `!isFolderImport` exclusion so folder imports follow the cursor. Kept both, so the resolved line is just `if (nextCursor && tweets.length > 0)` with no debug logging. The URL-construction rewrite merged cleanly. Green on my end.

ishaanxgupta (approved):

ishaanxgupta:
Thanks for the contribution!

PR #1270, opened 2026-07-12
fix(mcp): bound the API-key session fetch with a 30s timeout

abhay-codes07:
`validateOAuthToken` aborts its upstream `/v3/session` call after 30 seconds, but `validateApiKey` had no signal at all. A hung session response therefore stalled every API-key-authenticated MCP request until platform limits kicked in, instead of failing cleanly.

This applies the same `AbortSignal.timeout(30_000)` to the API-key session fetch. The existing catch already maps `TimeoutError` to a null auth result, so no other change is needed.

Note: an earlier revision of this PR also flushed PostHog events within the request lifetime; that was dropped at the maintainer's request since the buffering is intentional. The PR is now the single auth-timeout line.

### DIFF
--- a/apps/mcp/src/auth.ts
+++ b/apps/mcp/src/auth.ts
@@ -32,6 +32,7 @@ export async function validateApiKey(
 			headers: {
 				Authorization: `Bearer ${apiKey}`,
 			},
+			signal: AbortSignal.timeout(30_000),
 		})
 
 		if (!sessionResponse.ok) {


ishaanxgupta:
the 1st one only would happen when cloudfare worker stops, but we have kept it buffered for some reason so like could you change the PR just keeping it for 2nd?
@abhay-codes07

abhay-codes07:
Done. Dropped the PostHog flush commit and kept only the API-key auth fix, since the buffering is intentional on your side. The PR is now a single one-line change: `validateApiKey` adds `signal: AbortSignal.timeout(30_000)` on the `/v3/session` fetch, matching what `validateOAuthToken` already does, so a hung session response fails cleanly instead of stalling the request. The existing catch already maps `TimeoutError` to a null auth result, so no other change is needed. Retitled scope is just the auth timeout now.

ishaanxgupta:
Done!

PR #1277, opened 2026-07-12
fix(extension): avoid recapturing recalled memories

GautamSharma99:
### Summary

  When auto-search and auto-capture were enabled together, recalled memories were appended to the submitted prompt and then persisted through
  CAPTURE_PROMPT as if they were new user input.

  This caused retrieved memories to be repeatedly re-ingested, potentially amplifying duplicates, outdated information, and incorrect source
  attribution.

  ### Changes

  - Preserve the original user prompt before injecting recalled context.
  - Send only the original prompt to CAPTURE_PROMPT.
  - Continue injecting recalled memories into the composer for model context.
  - Apply the fix consistently to:
      - ChatGPT
      - Claude
      - T3

  - Preserve T3’s existing prompt-separation behavior.
  - Add shared prompt-preparation logic and regression tests.

  ### Testing

  - 4 prompt-capture regression tests pass.
  - Browser-extension TypeScript compilation passes.
  - Browser-extension production build passes.
  - Biome checks pass.
  - Full repository suite: 312 passed, 64 skipped, with 21 pre-existing unrelated failures.

  ### Notes

  This changes only the content persisted by auto-capture. The prompt sent to the chat platform still includes recalled memories as before.

### DIFF
--- a/apps/browser-extension/entrypoints/content/chatgpt.ts
+++ b/apps/browser-extension/entrypoints/content/chatgpt.ts
@@ -13,6 +13,7 @@ import {
 	createChatGPTInputBarElement,
 	DOMUtils,
 } from "../../utils/ui-components"
+import { preparePromptForSubmission } from "../../utils/prompt-capture"
 
 let chatGPTDebounceTimeout: NodeJS.Timeout | null = null
 let chatGPTRouteObserver: MutationObserver | null = null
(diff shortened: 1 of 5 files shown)


ishaanxgupta:
Hi @GautamSharma99 please fix the conflicts

ishaanxgupta:
Preventing recalled memories from being captured again is the right goal, but this implementation changes the existing capture flow and saves only the user prompt instead of the complete user/assistant conversation. In combined testing, it also interfered with recall reaching the model. This needs to be redesigned against current main while keeping recall injection and conversation capture independent. Closing this version for now

PR #1280, opened 2026-07-13
fix(cartesia): run the wrapped agent exactly once per event

abhay-codes07:
## What

`SupermemoryCartesiaAgent.process()` could run the wrapped agent **twice** for one event.

The whole method body sat in a single `try/except` whose fallback re-invoked `self.agent.process()`:

```python
try:
    ...memory enrichment...
    async for output in self.agent.process(env, event):
        yield output
except Exception as e:
    logger.error(f"[Supermemory] Error in process: {e}")
    async for output in self.agent.process(env, event):   # runs the WHOLE agent again
        yield output
```

The fallback exists so a memory failure never blocks the agent — but because the agent's own iteration lived inside the same `try`, an agent error raised **mid-stream** (LLM connection drop after chunks were already yielded) triggered the fallback too. The caller received a second, full response concatenated onto the partial one, plus a duplicate billable LLM call. This applied to every event type, not just `UserTurnEnded`.

## Fix

Scope the guard to the memory enrichment/injection/storage block only, then run the agent iteration exactly once after it:

- enrichment failures are logged and absorbed, and the agent still runs — same guarantee as before
- agent errors now propagate to the caller instead of triggering a duplicate run

No changes to the enrichment, prompt-injection, or storage logic themselves.

## Testing

New `tests/test_single_agent_run.py` streams from a fake inner agent that dies mid-stream:

- each chunk is delivered exactly once, the inner agent's run counter stays at 1, and the error propagates — **both mid-stream tests fail against the previous implementation** (verified: run counter hits 2 on the old code) 
- a third test pins the intended behaviour that an enrichment failure still lets the agent run exactly once

`python -m pytest tests` in the package — 4 passed.

cc @MaheshtheDev

### DIFF
--- a/packages/cartesia-sdk-python/src/supermemory_cartesia/agent.py
+++ b/packages/cartesia-sdk-python/src/supermemory_cartesia/agent.py
@@ -387,8 +387,14 @@ class SupermemoryCartesiaAgent:
         Yields:
             Output events from the wrapped agent.
         """
-        try:
-            if type(event).__name__ == "UserTurnEnded":
+        # Memory enrichment must never prevent the agent from running, so it
+        # gets its own guard. The agent iteration stays outside of it: with
+        # the old whole-body try/except, an agent error raised mid-stream
+        # (after some chunks were already yielded) triggered a fallback that
+        # re-ran the entire agent — duplicating the visible response and the
+        # LLM call.
+        if type(event).__name__ == "UserTurnEnded":
+            try:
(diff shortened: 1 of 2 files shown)


MaheshtheDev:
@abhay-codes07 can you fix the merge conflicts and please read the code and its quality why do we need 6 lines of comments?

abhay-codes07:
fixed it @MaheshtheDev I just mistakenly wrote the bug explaining in the code comment itself mistakenly not following the code structure

PR #1283, opened 2026-07-13
fix(web): register the drifted document endpoints in the api contract

abhay-codes07:
## What

The `@repo/lib` endpoint map and validation schemas have drifted behind what the web app actually exchanges with the API. Three concrete gaps:

1. **`categories` is missing from `DocumentsWithMemoriesQuerySchema`.** The memories grid and the X-bookmarks status query both filter by `categories` on `@post/documents/documents` — and both smuggle it past client validation with `disableValidation: true` because the schema doesn't know the field. Any future caller that filters by category *without* remembering that flag gets the key silently stripped and a filter that no-ops.

2. **`@post/documents/documents/facets` was never registered.** `DocumentFacetsQuerySchema` exists in the validation package but is imported nowhere — dead code — while the memories grid calls the endpoint as an untyped raw URL and hand-casts the response to a local type.

3. **`@patch/documents/:id` was never registered.** `MemoryUpdateSchema` exists for exactly this (its own comment pairs it with `MemoryAddSchema`, but only the add path got wired). The document-edit and file-metadata mutations PATCH via template-literal URLs (`` `@patch/documents/${id}` ``), bypassing all request/response typing.

## Changes

- `DocumentsWithMemoriesQuerySchema` gains `categories: z.array(z.string()).optional()`
- New `DocumentFacetSchema` / `DocumentFacetsResponseSchema` (mirroring the shape the grid already consumes, with `total` optional since the UI guards it), and `@post/documents/documents/facets` registered with the previously-dead query schema as input
- `@patch/documents/:id` registered with `MemoryUpdateSchema` / `MemoryResponseSchema` / `params`, and both PATCH call sites in `use-document-mutations.ts` converted to the typed route with `params: { id }`
- bun tests covering the `categories` field and the facets response shape

## Deliberately out of scope

The existing `disableValidation: true` flags stay put, including on the two calls that motivated the `categories` fix. Every current caller of `@post/documents/documents` sets the flag, which means output validation against `DocumentsWithMemoriesResponseSchema` has never been exercised in production — removing the flags safely needs someone to verify the declared response schemas against live responses (I don't have staging access). With the schemas now correct, that cleanup becomes a mechanical follow-up.

## Testing

- `bun test` in `packages/validation` — 12 pass (4 new)
- `tsc --noEmit -p packages/validation` — clean
- `bun test` + `bun run build` in `apps/web` — clean
- `biome check` on all touched files — clean

cc @MaheshtheDev

### DIFF
--- a/apps/web/hooks/use-document-mutations.ts
+++ b/apps/web/hooks/use-document-mutations.ts
@@ -522,7 +522,8 @@ export function useDocumentMutations({
 				const data = (await response.json()) as { id: string }
 
 				if (applyMeta && (title || description)) {
-					await $fetch(`@patch/documents/${data.id}`, {
+					await $fetch("@patch/documents/:id", {
+						params: { id: data.id },
 						body: {
 							metadata: {
 								...(title && { title }),
@@ -643,7 +644,8 @@ export function useDocumentMutations({
 			documentId: string
 			content: string
 		}) => {
-			const response = await $fetch(`@patch/documents/${documentId}`, {
+			const response = await $fetch("@patch/documents/:id", {
+				params: { id: documentId },
 				body: {
 					content,
 					metadata: { sm_source: "consumer" },
--- a/packages/lib/api.ts
+++ b/packages/lib/api.ts
@@ -11,6 +11,8 @@ import {
 	CreateProjectSchema,
 	DeleteProjectResponseSchema,
 	DeleteProjectSchema,
+	DocumentFacetsQuerySchema,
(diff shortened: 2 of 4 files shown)


abhay-codes07:
Rebased onto the latest `main` to clear the merge conflict. The only overlap was in `packages/validation/api.test.ts`, where `main` had since added a `pagination query schemas` block; kept both that and this PR's `documents contract schemas` block. `bun test` passes (31 tests). Ready for another look.

MaheshtheDev:

PR #1285, opened 2026-07-13
fix(tools): make claude-memory file operations act on the exact file, literally

abhay-codes07:
## What

Three related defects in the Claude memory tool's file operations (`packages/tools/src/claude-memory.ts`):

1. **`str_replace` corrupts content containing `$` sequences.** The model-authored replacement was passed to `String.replace` as a plain string, so JS replacement patterns were expanded: a `new_str` containing `$&`, `$'`, `` $` `` or `$$` silently wrote corrupted content while reporting success (replace `TODO` with `price is $& now` → stores `price is TODO now`). Memory files routinely contain literal dollar signs — code, prices, shell snippets.

2. **`view` could return the wrong file as a success.** `readFile` searched with `limit: 1` and then looked for the exact customId match *inside that single result* — while the sibling `getFileDocument` helper (used by every mutating command) uses `limit: 5` precisely so the exact match is findable among semantic near-neighbours. If a similarly-named file ranked first (`notes.txt` vs `notes_backup.txt`), `view` returned the wrong file's contents with line numbers, as a success.

3. **Both lookups fell back to `results[0]` when no exact match existed** — which let `view` read, and `str_replace`/`insert`/`delete`/`rename` *modify*, a different file than the one requested.

## Fix

- Function replacer keeps the replacement literal: `originalContent.replace(oldStr, () => newStr)`
- `readFile` now goes through `getFileDocument` (removing the duplicated lookup and making `view` read the same `raw || content` the mutating commands edit)
- The lookup requires the exact customId match and reports file-not-found otherwise — silently acting on a different file is strictly worse than an explicit error

## Testing

- New mocked-SDK tests: dollar-sequence literalness for all four patterns (`$&`, `$'`, `` $` ``, `$$`), exact-match preference when a neighbour ranks first, and not-found (with no write issued) when only a different file comes back — **six of the seven new tests fail against the previous implementation**
- `vitest run src/claude-memory.test.ts` — 10 pass
- `tsc --noEmit` error count unchanged vs `main` (148 pre-existing), `biome check` clean on touched files

cc @MaheshtheDev

### DIFF
--- a/packages/tools/package.json
+++ b/packages/tools/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@supermemory/tools",
   "type": "module",
-  "version": "2.1.0",
+  "version": "2.1.1",
   "description": "Memory tools for AI SDK, OpenAI, Voltagent and Mastra with supermemory",
   "scripts": {
     "build": "tsdown",
--- a/packages/tools/src/claude-memory.test.ts
+++ b/packages/tools/src/claude-memory.test.ts
@@ -4,12 +4,13 @@ import { beforeEach, describe, expect, it, vi } from "vitest"
 // can be exercised deterministically without any network access. We only need
 // `search.execute` to return a single document with known multi-line content.
 const searchExecute = vi.fn()
+const addMock = vi.fn()
 
 vi.mock("supermemory", () => {
 	return {
 		default: class MockSupermemory {
 			search = { execute: searchExecute }
-			add = vi.fn()
+			add = addMock
 			memories = { forget: vi.fn() }
 		},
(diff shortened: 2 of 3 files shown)


ishaanxgupta:
LGTM, @abhay-codes07 can you please bump the package?

ishaanxgupta (approved):

abhay-codes07:
Done, bumped to 2.1.1 in ebe34679. Thanks for the review @ishaanxgupta.

ishaanxgupta (approved):

PR #1294, opened 2026-07-14
ci: grant least-privilege GITHUB_TOKEN in ci.yml

### LINKED ISSUE #1218: CI workflow (ci.yml) runs without a least-privilege GITHUB_TOKEN
## Problem

The `.github/workflows/ci.yml` workflow does not declare a `permissions:` block. Without one, the injected `GITHUB_TOKEN` inherits the repository or organization default, which is commonly broad read/write.

## Motivation

The `quality-checks` job only needs to read the repository (checkout, install dependencies, type check, Biome). Granting it a write-capable token is unnecessary and widens the blast radius: a compromised step or a malicious dependency could use that token to push commits, alter releases, or otherwise move laterally. Setting a least-privilege default is a standard CI hardening step and matches the pattern already used by the `publish-*.yml` workflows in this repo.

## Expected behavior

The CI workflow should run with a read-only token by default, and only request additional scopes if a specific step needs them.

## Possible implementation

Add a top-level read-only default to `ci.yml`:

```yaml
on:
  pull_request:

permissions:
  contents: read

jobs:
  quality-checks:
    ...
```

0xkaushik-ai:
## Summary

- Adds an explicit workflow-level `permissions: contents: read` block to `.github/workflows/ci.yml`.
- The `quality-checks` job only checks out the repo and runs typecheck/lint, so it does not need write access.
- Matches the least-privilege pattern already used by the publish workflows and addresses the supply-chain / blast-radius concern in #1218.

## Test plan

- [x] Confirmed the workflow YAML is valid and only changes permissions.
- [ ] CI runs successfully on this PR with the restricted token (checkout, install, typecheck, Biome).

Fixes #1218

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -3,6 +3,11 @@ name: CI - Type Check, Format & Lint
 on:
   pull_request:
 
+# Least-privilege token: this workflow only checks out the repo and runs
+# typecheck/lint. It does not need write access.
+permissions:
+  contents: read
+
 concurrency:
   group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
   cancel-in-progress: true


ishaanxgupta:
Not required as of now! thanks for the contribution @shitcodebykaushik

PR #1308, opened 2026-07-18
feat(memory-graph): add in-graph node search

abhay-codes07:
## What

Adds node search to the memory graph. The graph already supports pan, zoom, fit, center, and arrow-key navigation between neighbours, but there was no way to jump to a specific document or memory by name. On a graph with a few hundred nodes, finding "the Berlin memory" meant scanning the canvas by eye.

This adds a search box to the bottom-left control stack that finds nodes by text, highlights where the matches are, and steps through them.

## How it works

- **`searchNodes(nodes, query)`** — a pure, case-insensitive substring matcher over document `title`/`summary`/`type` and memory `memory`/`content`. Matches are ranked (primary-field prefix > primary-field substring > supporting-copy substring), with the input order as a stable tiebreaker so results never depend on where the force simulation dropped a node. Exported from the package for advanced/headless use.
- **`SearchControl`** — an input styled to match the existing navigation controls, with a live `current/total` counter, prev/next steppers, and a clear button.
- **Wiring in `MemoryGraph`** — the focused match is selected and centered through the same `selectAndCenter` path arrow-key navigation already uses, and the documents behind every match are highlighted (a memory match highlights its parent document) by merging into the existing `highlightDocumentIds` pipeline rather than adding a parallel one.

## Interaction

- `/` focuses the search box, consistent with the existing single-key shortcuts (`z` fit, `c` center, `+`/`-` zoom)
- `Enter` / `Shift+Enter` step to the next / previous match
- `Escape` clears the search
- The existing keyboard handlers already ignore events originating from inputs, so typing a query never triggers the fit/zoom/arrow shortcuts

## Configuration

On by default. Embedders can hide it with `showSearch={false}` or relabel the placeholder with `searchPlaceholder`. I defaulted it on because it is the point of the PR and it degrades gracefully (it renders only when there are nodes), but if you would rather ship it opt-in I am happy to flip the default.

## Testing

- `searchNodes`: empty/whitespace query, case-insensitivity, memory text vs content matching, primary-over-secondary ranking, prefix-over-substring ranking, stable order for ties, null-title/missing-content tolerance
- `SearchControl` (mounted): typing forwards to the handler, counter shows 1-based position and total, zero-results state, `Enter`/`Shift+Enter` stepping, `Escape` and clear-button behaviour, disabled steppers with no matches
- 14 new tests; full package suite green (202 passing), `tsc --noEmit` clean, `biome check` clean, `vite build` succeeds

cc @MaheshtheDev @ishaanxgupta

### DIFF
--- a/packages/memory-graph/README.md
+++ b/packages/memory-graph/README.md
@@ -51,6 +51,7 @@ function App() {
 - **Interactive canvas visualization** - Pan, zoom, and drag nodes using Canvas 2D rendering
 - **Document and memory nodes** - Documents as rectangles, memories as hexagons
 - **Relationship visualization** - Edges show document similarity and memory version chains
+- **Node search** - Find documents and memories by text, highlight matches, and step through them (press `/` to focus, `Enter` / `Shift+Enter` to cycle)
 - **Space filtering** - Filter by workspace or view all memories
 - **Two variants** - Full-featured console mode or embedded consumer mode
 - **Pagination support** - Load more documents on demand
@@ -66,6 +67,8 @@ function App() {
 | `error` | `Error \| null` | Error to display |
 | `loadMoreDocuments` | `() => Promise<void>` | Function to load more data |
 | `highlightDocumentIds` | `string[]` | IDs of documents to highlight |
+| `showSearch` | `boolean` | Show the in-graph node search box (default: `true`) |
+| `searchPlaceholder` | `string` | Placeholder text for the search box |
 
 ## Documentation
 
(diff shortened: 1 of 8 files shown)


ishaanxgupta:
Hi @abhay-codes07 this feature has not been planned yet, although I think this could be a great addition, will discuss with team and let you know, till then changing the status to draft

abhay-codes07:
Sounds good, thanks for taking a look. Happy to leave it in draft while the team decides whether in-graph search fits the roadmap. The feature is behind a `showSearch` prop that defaults on, so it is easy to flip to opt-in or drop entirely depending on what you land on. Ping me when there is a direction and I will rebase and adjust.

PR #1326, opened 2026-07-21
fix(pipecat): use client.add, memories.add was removed in supermemory 3.x

Cintu07:
supermemory_pipecat stores messages with `client.memories.add(...)`, but
`.add` was moved off `memories` to the top level (`client.add(...)`) in
supermemory 3.x. the package pins `supermemory>=3.16.0`, a range where
`memories.add` no longer exists, so a fresh install pulls 3.51 and every
store raises AttributeError.

it's worse than a hard crash: the call sits in a fire-and-forget block that
only logs the error (service.py _store_messages), so memory silently never
gets stored and the user never sees why.

fix is one line: call `client.add(**add_params)`. confirmed the top-level
add accepts the same params this passes (content, container_tags, custom_id,
metadata), and that `client.memories.add` is gone on current supermemory ^^

### DIFF
--- a/packages/pipecat-sdk-python/src/supermemory_pipecat/service.py
+++ b/packages/pipecat-sdk-python/src/supermemory_pipecat/service.py
@@ -179,7 +179,7 @@ class SupermemoryPipecatService(FrameProcessor):
             if self.session_id:
                 add_params["custom_id"] = self.session_id
 
-            await self._supermemory_client.memories.add(**add_params)
+            await self._supermemory_client.add(**add_params)
 
         except Exception as e:
             logger.error(f"Error storing messages: {e}")
--- a/packages/pipecat-sdk-python/tests/test_empty_profile.py
+++ b/packages/pipecat-sdk-python/tests/test_empty_profile.py
@@ -121,3 +121,29 @@ class TestSupermemoryPipecatNullProfile(unittest.IsolatedAsyncioTestCase):
                 "search_results": [],
             },
         )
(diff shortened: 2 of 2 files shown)


Cintu07:
same root cause as #1236 (supermemory 3.x moved .add off .memories), just in
a different sdk. that one's an install-time cap, this one's a runtime call
that fails silently. this is very small and independent fix. so yeah

Cintu07:
this fixes a silent bug: pipecat stores with client.memories.add, which supermemory 3.x removed, so every save quietly fails inside the fire-and-forget block and the user never sees it. one line fix to client.add.

also added the regression test copilot suggested: it asserts _store_messages calls client.add and not the old memories.add, so this cannot silently break again. both tests pass.

the bug is still on main and this has only had the bot look at it so far. mind giving it a review when you get a sec @ishaanxgupta?

Cintu07:
Bumping this. It is the same supermemory 3.x change that #1236 (just merged) worked around: memories.add moved off the client, so pipecat, which pins supermemory >=3.16, raises AttributeError on every store today. The call sits in a fire-and-forget block that only logs, so memory silently never stores and the user never sees why. The fix is the one-line swap to client.add, and I checked it takes the same params. @ishaanxgupta could you take a look when you get a chance?

MaheshtheDev:
Already fixed on main. Pipecat already calls `client.add()` (the v4 API). Closing as already landed.

PR #1328, opened 2026-07-21
fix(mcp): point metadata to human docs

DisturbedCrow:
Fixes the MCP metadata links so they point to the public MCP guide instead of the protocol endpoint.

Biome passes and the target URL returns 200.


ishaanxgupta:
Done!

PR #1339, opened 2026-07-22
fix(extension): stop fragmenting Included Memories that contain commas or newlines

abhay-codes07:
Follow-up to #1261, rebuilt to stay within the current architecture. That PR was closed because it changed the shared `GET_RELATED_MEMORIES` response format and regressed auto-recall. This version does not touch that contract at all. It fixes only the one data-handling bug from #1257 that survived the `memory-suggestion.ts` refactor, and it does so entirely inside the popup's own display data.

### The bug

The Included Memories popup stores the memory list on a `data-*` attribute by joining it into one string, then reads it back by splitting on `[,\n]`:

- store: `iconElement.dataset.memoriesData = String(response.data)` (chatgpt.ts, claude.ts, gemini.ts) and `= response.data` (t3.ts)
- read: `memories.split(/[,\n]/)` in `showMarkerPopover` (memory-suggestion.ts) and in T3's popup list and delete handler

Memory text is free-form and regularly contains commas and newlines, so one memory gets shredded into several rows. For example "Lives in Austin, Texas" shows as two entries. In T3 it is worse than cosmetic: the per-item delete indexes into that split list, so removing an entry splices the wrong item and then rewrites the injected prompt from the mangled result.

### The fix

Store the list as JSON on the attribute and parse it back, so a memory with a comma or newline round-trips as a single item:

- `serializeMemoriesForDataset(memories)` — JSON, returns `""` for an empty list so the existing "are there memories" truthiness checks on the attribute keep working unchanged.
- `parseMemoriesFromDataset(raw)` — parses the JSON, and falls back to the legacy `[,\n]` split for any value written by older code, so nothing breaks on upgrade.

Applied consistently across all four consumers (ChatGPT, Claude, Gemini, T3).

### Why this preserves the contract

The change is confined to `dataset.memoriesData`, which only feeds the popup UI. The `GET_RELATED_MEMORIES` message is untouched (still an array of raw strings), and the injected `dataset.supermemories` prompt text is byte-for-byte the same as before (T3's initial injection still uses `${response.data}`, and its delete-path re-injection still uses the same joined form). So what the model sees, and therefore recall behaviour, does not change.

### Tests

`memory-suggestion.test.ts` (bun:test, following `search-request.test.ts`) covers the round-trip: comma-bearing and newline-bearing memories stay single items, trimming/empty handling, empty-list serializes to `""`, and the legacy-split fallback. 7 tests pass. Extension `tsc` is clean for these files (the remaining `new_logo.png` `getURL` overload errors are pre-existing on `main`). cc @ishaanxgupta


HaoChiBao:
## Coordination

Taking the leftover work from #1257 that this PR intentionally left out (per your description — data-handling / comma split only).

Specifically (T3 path on `main`):
- `<= 1` wipe guard → `=== 0`
- dispose previous popup + document click listener before mounting a new one
- rebuild injected `dataset.supermemories` from the remaining memories without `" ,"` / stale numbering

I'll open a follow-up PR based on this branch so we don't fight over the JSON serialize helpers. Ping me if you'd rather fold any of that into #1339 instead.

HaoChiBao:
﻿Follow-up PR opened: https://github.com/supermemoryai/supermemory/pull/1421

Includes the #1339 comma-split commit plus T3 wipe/listener/prompt renumber fixes for bugs 2–4.

ved015 (approved):
LGTM !

ishaanxgupta (approved):

PR #1340, opened 2026-07-22
fix(web): stop formatUsageNumber rendering 1000.0K at unit boundaries

abhay-codes07:
### Problem

`formatUsageNumber` scales the value and formats it with one decimal:

```ts
if (value >= 1_000) {
  const thousands = value / 1_000
  return thousands % 1 === 0 ? `${thousands}K` : `${thousands.toFixed(1)}K`
}
```

`toFixed(1)` rounds, so any value from 999,950 to 999,999 becomes `999.95..999.999` → `"1000.0"`, and the usage counter shows **"1000.0K"** instead of "1.0M". The same artifact appears at the million boundary (999,999,950 → "1000.0M").

### Fix

When the scaled value rounds up to 1000, promote it to the next unit, and add a billions tier so very large counts read "1.0B" instead of "1000.0M":

```ts
if (value >= 1_000) {
  const thousands = value / 1_000
  return thousands >= 999.95
    ? withSuffix(value / 1_000_000, "M")
    : withSuffix(thousands, "K")
}
```

### Tests

Added `billing-utils.test.ts` (bun:test) covering both boundaries: 999,950 and 999,999 → "1.0M", 999,949 → "999.9K" (still rounds within band), 999,999,950 → "1.0B", plus the existing 50K / 1.5M / 2M cases. 5 tests pass.


ved015 (approved):
LGTM!

ishaanxgupta (approved):

PR #1345, opened 2026-07-23
fix(mcp): require confirmation before forgetting similar memories

DisturbedCrow:
Stops `forget` from immediately deleting a semantic near-match. It now previews the candidate and requires a short-lived signed confirmation.

Focused tests and the Worker dry-run build pass.


ishaanxgupta:
New MCP is out, this is already solved!

PR #1421, opened 2026-08-06
fix(extension): finish Included Memories leftovers on T3 (#1257)

### LINKED ISSUE #1257: Extension: Included Memories popup fragments comma-containing memories, leaks global click listeners, and drops the last memory on removal

HaoChiBao:
## Summary
- Completes the leftover bugs from #1257 that [#1339](https://github.com/supermemoryai/supermemory/pull/1339) intentionally left for a follow-up (T3 Included Memories path).
- Includes @abhay-codes07's JSON `dataset.memoriesData` round-trip (cherry-picked from #1339) so comma-bearing memories and delete indexes stay correct.
- Fixes T3 wipe-on-remove (`<= 1` → only clear when empty), disposes prior popup + `document` click listeners before remounting, and rebuilds the injected prompt with renumbered memories via `buildSupermemoryText`.

## Why this PR includes #1339
Bugs 2–4 depend on correct list indexing. Landing only the lifecycle fix on `main` without the JSON serialize would keep delete indexes wrong for comma-containing memories. Happy to adjust if maintainers prefer to merge #1339 first and drop that commit from this PR.

## Test plan
- [x] `bun test entrypoints/content/memory-suggestion.test.ts` (9 pass)
- [ ] On t3.chat: trigger Included Memories with 2+ memories; remove one → the other remains injected
- [ ] Trigger auto-search repeatedly → only one popup / click listener remains (no stacked body nodes)
- [ ] Remove all memories → marker resets; leave one → prompt renumbers `1. …`

Fixes #1257

Related: #1339


ved015:
Hi @HaoChiBao thanks for the contribution favoured pr #1339 as it was opened earlier please fix the conflicts and then tag me i will have a review for the rest of the changes

Thanks !

HaoChiBao:
Hey @ved015 thanks for taking a look!

Conflicts are fixed now. I merged latest `main` (including #1339) into this branch and kept just the T3 follow-up bits on top: the empty-only wipe, disposing the old popup/click listener before remounting, and rebuilding the injected prompt with renumbered memories.

Happy to adjust anything once you've had a chance to review.

ved015 (approved):

ishaanxgupta (approved):

PR #1427, opened 2026-08-07
fix(openai-sdk): migrate to supermemory 3.x api so a fresh install imports

### LINKED ISSUE #1235: supermemory-openai-sdk fails to import with the latest supermemory (3.50)

Cintu07:
fixes #1235.

a fresh pip install supermemory-openai-sdk pulls the latest supermemory and the import fails. supermemory 3.x renamed the types the sdk imports and moved add off client.memories.

this moves the sdk onto the 3.x api:

- imports: MemoryAddResponse -> AddResponse, MemoryGetResponse -> DocumentGetResponse
- calls: client.memories.add(...) -> client.add(...) in tools.py and middleware.py
- deps: supermemory>=3.7.0, version 1.0.5

verified the package imports on the declared floor supermemory 3.7.0 (the first release with client.add and AddResponse; 3.5.0 and 3.6.0 do not have them) and on the 3.50.0 line, and the test suite runs again (24 passed, 11 skipped) where it could not even be collected before.

### DIFF
--- a/packages/openai-sdk-python/pyproject.toml
+++ b/packages/openai-sdk-python/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "hatchling.build"
 
 [project]
 name = "supermemory-openai-sdk"
-version = "1.0.4"
+version = "1.0.5"
 description = "Memory tools for OpenAI function calling with supermemory"
 readme = "README.md"
 license = "MIT"
@@ -26,7 +26,7 @@ classifiers = [
 requires-python = ">=3.8.1"
 dependencies = [
     "openai>=1.102.0",
-    "supermemory>=3.1.0",
+    "supermemory>=3.5.0",
     "typing-extensions>=4.0.0",
     "requests>=2.25.0",
 ]
--- a/packages/openai-sdk-python/pyproject.toml
+++ b/packages/openai-sdk-python/pyproject.toml
@@ -26,7 +26,7 @@ classifiers = [
 requires-python = ">=3.8.1"
 dependencies = [
     "openai>=1.102.0",
-    "supermemory>=3.5.0",
+    "supermemory>=3.7.0",
     "typing-extensions>=4.0.0",
     "requests>=2.25.0",
 ]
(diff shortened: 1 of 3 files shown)


Cintu07:
heads up, since this makes the openai-sdk tests collectable again, it should unblock adding this package to CI. #1417 mentions the openai-sdk part was waiting on this issue getting fixed

Cintu07:
dam that was good catch on dependency floor, anyway fixed: bumped to supermemory>=3.7.0, which is the first version that actually has client.add and AddResponse (checked 3.5 and 3.6 don't have them, 3.7 does).

the memory type mismatch is pre-existing, it was there before this pr under the old MemoryAddResponse name, so leaving it out of scope to keep this a clean api migration. happy to do it as a follow up hehe ^^

Cintu07:
closing this in favor of #1236, which shipped the version cap and is now merged. that already fixes the fresh-install import break this pr was targeting, so the full migration here is redundant. happy to revisit if you decide to move the sdk to the 3.x api later.

PR #1438, opened 2026-08-08
fix(python-sdks): handle pydantic search results in memory dedup helpers

### LINKED ISSUE #1266: Python SDK packages crash or silently drop memories: dedup helpers treat pydantic search results as dicts (pipecat, cartesia, agent-framework)
**Environment:** current `main`; affects `supermemory-pipecat`, `supermemory-cartesia`, and `supermemory-agent-framework` (all three Python integration packages that fetch memories through the `supermemory` SDK).

The three SDK-based Python packages pass search results into dedup/format helpers that were written for plain dicts, but the SDK returns **pydantic models** (attribute access, snake_case fields). The result is that memory injection — the core feature of these packages — is broken on its primary path:

**pipecat** (`src/supermemory_pipecat/utils.py`): `deduplicate_memories`'s `unique_search` calls `r.get("memory", "")` on a `Result` model → `AttributeError: 'Result' object has no attribute 'get'`. The exception escapes `_enhance_context_with_memories`, is caught by the generic handler in `process_frame` ("Error processing frame"), and the frame is forwarded unchanged — so **no memories are ever injected for any user whose profile lookup returns search results**. It only appears to work for brand-new users with empty search results.

**cartesia** (`src/supermemory_cartesia/utils.py`): identical helper, identical crash. Here it's swallowed by `_enrich_event_with_memories`'s `except Exception` which returns `(event, None)` — and the process method then strips the previous memory block from the system prompt and injects nothing. Persistent recall is effectively dead for users with stored memories.

**agent-framework** (`src/supermemory_agent_framework/utils.py`): `extract_memory_text` handles only `dict` and `str`, so pydantic items fall through to `return None` — every search-result memory is **silently dropped** from the injected context (no crash, just data loss).

The asymmetry that hid this: the OpenAI package shares the same helper shape but fetches profile data over raw `aiohttp` and receives JSON **dicts**, so it works.

**Reproduction (no API key needed)** — build a real SDK model and feed it through the pipecat helper:

```python
from supermemory.types.search_memories_response import Result
from supermemory_pipecat.utils import deduplicate_memories

r = Result.model_validate({
    "id": "mem_1", "similarity": 0.9,
    "memory": "User prefers async", "updatedAt": "2026-01-01T00:00:00Z",
})
deduplicate_memories(static=[], dynamic=[], search_results=[r])
# AttributeError: 'Result' object has no attribute 'get'
```

Fix is ready (shared field-extraction that handles dicts with camelCase keys and models with snake_case attributes, plus regression tests in all three packages) — PR incoming.

One adjacent note from testing: the agent-framework package no longer imports against any installable `agent-framework-core` release — its pin is `>=1.0.0rc3`, but `BaseContextProvider` was removed from the package after the rc series (1.0.0 stable through 1.11.0 don't export it), so `pip install supermemory-agent-framework` today yields an ImportError on import. That likely deserves its own issue/fix.

vivekvar-dl:
Fixes #1266

## Problem

The three SDK-based Python packages pass search results into dedup/format helpers written for plain dicts, but typed SDK results are pydantic models — attribute access, snake_case fields, no `.get()`:

- **pipecat** / **cartesia**: `deduplicate_memories`'s `unique_search` calls `r.get("memory", "")` → `AttributeError` on a model. The exception is swallowed upstream (generic frame/event handlers), so **no memories are injected for any user whose lookup returns search results** — it only appears to work for brand-new users.
- **agent-framework**: `extract_memory_text` handles only `dict` and `str`, so model items fall through to `None` and every search-result memory is **silently dropped**.

A version wrinkle makes this worse: on current `supermemory` (3.56.0) the profile endpoint types `search_results.results` as `List[object]`, so items arrive as **camelCase dicts** — while typed models (e.g. `search_memories_response.Result`) expose **snake_case attributes**. The packages pin `supermemory>=3.16.0`, so both shapes occur in the wild and the helpers must accept both.

## Fix

Implements the shared field-extraction approach described in the issue:

- **pipecat + cartesia `utils.py`** (kept byte-identical, matching the existing duplication between the two packages): a private `_get_result_field(result, *keys)` reads a field from either shape — dict lookup with camelCase/snake_case keys, or attribute access. Used in `unique_search` and `format_memories_to_text`. Plain-string items and camelCase dicts behave exactly as before; `updated_at`/`updatedAt` both resolve for the timestamp.
- **agent-framework `utils.py`**: `extract_memory_text` gains an attribute-access fallback for non-dict, non-str items. Dict/str/None behavior is unchanged.

## Tests

- New `tests/test_utils.py` in pipecat and cartesia (unittest style, matching the existing suite): model items survive dedup, dedup against profile strings, camelCase dicts unchanged, timestamp formatting for both shapes, plain strings verbatim. The dependency stubs moved from `test_empty_profile.py` into a shared `tests/conftest.py` so both test modules reuse them.
- Three new cases in agent-framework's existing `TestDeduplicateMemories` (pytest style): model items extracted, deduped against static, empty/None memory filtered.

All three suites pass locally (Python 3.12): pipecat 9/9, cartesia 9/9, agent-framework 57/57 (with `agent-framework-core==1.0.0rc3` — the current stable releases break the package's import, tracked separately per the issue's closing note).

Also verified the issue's exact reproduction against the real SDK (`supermemory` 3.56.0): `search_memories_response.Result` now dedups and formats correctly in all three packages, and camelCase dict inputs are unchanged.

### DIFF
--- a/packages/agent-framework-python/src/supermemory_agent_framework/utils.py
+++ b/packages/agent-framework-python/src/supermemory_agent_framework/utils.py
@@ -92,14 +92,15 @@ def deduplicate_memories(
     def extract_memory_text(item: Any) -> Optional[str]:
         if item is None:
             return None
-        if isinstance(item, dict):
-            memory = item.get("memory")
-            if isinstance(memory, str):
-                trimmed = memory.strip()
-                return trimmed if trimmed else None
-            return None
         if isinstance(item, str):
-            trimmed = item.strip()
+            memory: Any = item
+        elif isinstance(item, dict):
+            memory = item.get("memory")
+        else:
+            # SDK search results are pydantic models with a `memory` attribute
+            memory = getattr(item, "memory", None)
+        if isinstance(memory, str):
+            trimmed = memory.strip()
             return trimmed if trimmed else None
         return None
 
--- a/packages/agent-framework-python/tests/test_utils.py
+++ b/packages/agent-framework-python/tests/test_utils.py
@@ -1,5 +1,7 @@
 """Tests for utility functions."""
 
+from types import SimpleNamespace
+
 import pytest
 
 from supermemory_agent_framework.utils import (
@@ -55,6 +57,30 @@ class TestDeduplicateMemories:
             static=[None, {"memory": "valid"}],
         )
         assert result.static == ["valid"]
+
+    def test_model_items(self) -> None:
+        # SDK search results are pydantic models with a `memory` attribute,
+        # not dicts — they must not be silently dropped (#1266).
+        result = deduplicate_memories(
+            search_results=[SimpleNamespace(memory="User prefers async")],
+        )
+        assert result.search_results == ["User prefers async"]
+
+    def test_model_items_deduplicate_against_static(self) -> None:
+        result = deduplicate_memories(
+            static=[{"memory": "User likes Python"}],
+            search_results=[
+                SimpleNamespace(memory="User likes Python"),
+                SimpleNamespace(memory="User prefers async"),
+            ],
+        )
+        assert result.search_results == ["User prefers async"]
+
+    def test_model_items_without_memory_text_filtered(self) -> None:
+        result = deduplicate_memories(
+            search_results=[SimpleNamespace(memory=None), SimpleNamespace(memory="  ")],
+        )
+        assert result.search_results == []
 
 
 class TestConvertProfileToMarkdown:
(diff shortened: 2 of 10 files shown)


MaheshtheDev:
Already fixed on main in #1434. Agent-framework, Cartesia, and Pipecat already read search results from both dicts and SDK models. Closing as already landed.

PR #1445, opened 2026-08-08
fix: stop uploading Sentry sourcemaps after builds

### LINKED ISSUE #1444: fix(build): do not run Sentry release upload as a root postbuild hook
## Summary

The root `build` command automatically runs a Sentry release creation and sourcemap upload after `turbo run build`.

This makes normal local and CI builds depend on Sentry credentials and attempts to upload a root-level `dist` directory that the Turbo build does not produce.

## Current behavior

`package.json`:

```json
{
  "scripts": {
    "build": "turbo run build",
    "sentry:sourcemaps": "_SENTRY_RELEASE=$(sentry-cli releases propose-version) && sentry-cli releases new $_SENTRY_RELEASE --org=supermemory --project=consumer-app && sentry-cli sourcemaps upload --org=supermemory --project=consumer-app --release=$_SENTRY_RELEASE --strip-prefix 'dist/..' dist",
    "postbuild": "bun run sentry:sourcemaps"
  }
}
```

Running:

```bash
bun run build
```

first completes the Turbo build, then invokes `sentry:sourcemaps`.

## Problem

- Local builds fail unless Sentry authentication/configuration is available.
- CI build validation also requires deployment-only credentials.
- The upload script targets `dist`, while application builds such as Next.js produce `.next` output.
- The Sentry command uses POSIX shell syntax, which may not be portable to Windows environments.

## Expected behavior

`bun run build` should only build and validate the repository. Sentry releases and sourcemap uploads should run only in an explicit deployment/release workflow, with the correct application output path.

## Proposed fix

- Remove the root `postbuild` hook.
- Keep Sentry upload as an explicit command, e.g. `bun run sentry:sourcemaps`.
- Invoke that command only from the production deployment pipeline after the relevant app build.
- Update the sourcemap path and `--strip-prefix` to match the deployed app’s actual build output.
- Prefer a cross-platform environment-variable approach or run the release step in the CI shell that owns deployment.

## Acceptance criteria

- `bun run build` succeeds without `SENTRY_AUTH_TOKEN` or other Sentry configuration.
- No Sentry release/upload is attempted during ordinary builds.
- Production deployment still creates the intended Sentry release and uploads valid sourcemaps.

abhinavkr26104:
Closes #1444 

## Summary

- remove the root `postbuild` hook that automatically ran Sentry sourcemap publishing
- keep `bun run build` limited to `turbo run build`
- retain `bun run sentry:sourcemaps` as an explicit release-only command

## Why

Ordinary local and CI builds should not require Sentry credentials or attempt release creation/sourcemap uploads.

## Validation

- verified `package.json` parses successfully
- verified the diff has no whitespace errors

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,8 +7,7 @@
     "dev:local": "turbo run dev:app",
     "format-lint": "bunx biome check --write",
     "check-types": "turbo run check-types",
-    "sentry:sourcemaps": "_SENTRY_RELEASE=$(sentry-cli releases propose-version) && sentry-cli releases new $_SENTRY_RELEASE --org=supermemory --project=consumer-app && sentry-cli sourcemaps upload --org=supermemory --project=consumer-app --release=$_SENTRY_RELEASE --strip-prefix 'dist/..' dist",
-    "postbuild": "bun run sentry:sourcemaps"
+    "sentry:sourcemaps": "_SENTRY_RELEASE=$(sentry-cli releases propose-version) && sentry-cli releases new $_SENTRY_RELEASE --org=supermemory --project=consumer-app && sentry-cli sourcemaps upload --org=supermemory --project=consumer-app --release=$_SENTRY_RELEASE --strip-prefix 'dist/..' dist"
   },
   "engines": {
     "node": ">=20"


MaheshtheDev:

Issue #1446, opened 2026-08-08
fix: make check-types cover all major workspaces


abhinavkr26104:
## Problem

`bun run check-types` runs `turbo run check-types`, but several major workspaces do not define a `check-types` script. Turbo therefore skips them silently.

Affected workspaces include:

- `apps/web`
- docs
- browser extension
- UI
- hooks
- lib
- validation

This makes the documented repository-wide type-check command incomplete.

## Expected behavior

`bun run check-types` should type-check every supported workspace, or the documentation/command should explicitly state its limited scope.

## Suggested fix

Add appropriate `check-types` scripts to each workspace, typically using `tsc --noEmit` (or the framework-specific equivalent), then verify the root Turbo task discovers and runs them.

Issue #1249, opened 2026-07-12
`turbo run check-types` is a no-op — no workspace package implements the script


rajarshidattapy:
## Summary

`check-types` is wired in root + turbo but no workspace package implements it — CI type-checking is silently skipped

## Labels

`bug`, `dx`, `ci`

---

### Summary

The root `package.json` defines:

```json
"check-types": "turbo run check-types"
```

and `turbo.json` has a pipeline entry for it:

```json
"check-types": {
    "dependsOn": ["^check-types"]
}
```

However, **no workspace package** (`apps/*`, `packages/*`) defines a `check-types` script in its own `package.json`. Turbo silently skips packages that don't declare the requested script, so `bun run check-types` / `turbo run check-types` completes successfully while checking **zero** TypeScript code.

### Affected packages

| Package | Has `check-types`? | Notes |
|---|---|---|
| `apps/mcp` | ❌ | Has a valid `tsconfig.json` with `strict: true` |
| `apps/web` | ❌ | |
| `apps/docs` | ❌ | |
| `apps/browser-extension` | ❌ | Has `"compile": "tsc --noEmit"` but not named `check-types` |
| `apps/memory-graph-playground` | ❌ | |
| `packages/lib` | ❌ | |
| `packages/validation` | ❌ | |

### Impact

- Type errors can be merged without CI catching them.
- Contributors may assume `bun run check-types` validates the entire monorepo when it actually validates nothing.
- `apps/browser-extension` already has the right command (`tsc --noEmit`) but under a different script name (`compile`), so it's also missed by Turbo.

### Suggested fix

Add a `check-types` script to every workspace package that has a `tsconfig.json`:

```json
"check-types": "tsc --noEmit"
```

For `apps/browser-extension`, rename or alias the existing `compile` script:

```json
"check-types": "tsc --noEmit"
```

Optionally, if some packages don't have `tsc` as a direct dependency, a shared approach could be added (e.g., using the root TypeScript installation via `turbo`'s `passThroughEnv` or a shared script).

### How I found this

Ran `tsc` manually against `apps/mcp/tsconfig.json` and noticed real type errors. Checked why CI didn't catch them and discovered the root `check-types` script was silently succeeding because no package implements it.

PR #1457, opened 2026-08-12
fix(web): stop NEXT_PUBLIC_*_AUTH_ENABLED from hiding the provider it names

### LINKED ISSUE #1278: Login page: NEXT_PUBLIC_*_AUTH_ENABLED flags act inverted (setting them hides the OAuth buttons)
**Environment:** apps/web on current `main`, login page (`app/(auth)/login/page.tsx`), lines ~356 and ~413.

The Google and GitHub buttons on the login page are gated like this:

```tsx
{process.env.NEXT_PUBLIC_HOST_ID === "supermemory" ||
 !process.env.NEXT_PUBLIC_GOOGLE_AUTH_ENABLED ? ( /* Google button */ )
```

Note the `!`. Taken at face value this means: on a non-`supermemory` host, setting `NEXT_PUBLIC_GOOGLE_AUTH_ENABLED=true` **hides** the Google button, and leaving it unset **shows** it. Same for GitHub. That directly contradicts the `_AUTH_ENABLED` name.

These two lines are the only references to either variable in the entire repo — no `.env.example` entry, no docs — so I can't tell which behavior is intended:

- **If the intent is "show the button when the provider is enabled"** (what the name says), the `!` should be dropped. But that flips the default for self-hosted deployments: anyone currently relying on the buttons appearing without setting the flag would see them disappear.
- **If the intent is "show by default, allow opting out"**, the current logic is right but the variable is badly misnamed — it behaves as `..._AUTH_DISABLED` — and deserves a rename or at least a comment, because the first self-hoster who sets `GOOGLE_AUTH_ENABLED=true` to turn Google on will get the opposite.

Either resolution is a one-line change; flagging it as an issue rather than a PR since the right default for self-hosted deployments is a product decision. Happy to send the patch once the intended semantics are confirmed.

Chirag6722:
Fixes #1278.

## The bug

The Google and GitHub buttons on the login page were gated like this:

```tsx
{process.env.NEXT_PUBLIC_HOST_ID === "supermemory" ||
 !process.env.NEXT_PUBLIC_GOOGLE_AUTH_ENABLED ? ( /* Google button */ )
```

Because of the `!`, a self-hosted deployment that set `NEXT_PUBLIC_GOOGLE_AUTH_ENABLED=true` to turn Google on got the opposite result — the button disappeared. Same for GitHub.

## Why not just drop the `!`

#1278 deliberately stopped short of a patch, because the two readings imply different defaults and the right one is a product call:

- **Drop the `!`** (opt-in): matches the name, but the buttons vanish for every existing self-hoster who never set the flag. Breaking.
- **Keep current behaviour** (opt-out): the default is fine, but the variable is misnamed and setting it to `true` does the wrong thing.

This PR takes the second reading and makes the implementation honest, so no existing deployment changes behaviour:

| flag value | before | after |
|---|---|---|
| unset | shown | shown (unchanged) |
| `true` | **hidden** ← the bug | shown |
| `false` / `0` | shown | hidden (new — a real off-switch) |

Only the previously broken case changes. The flag now never hides the provider it names, and self-hosters get an actual way to turn a provider off, which the old logic couldn't express either.

If you'd rather have true opt-in semantics (unset = hidden), say so and I'll flip it — it's a two-line change to `isSocialProviderEnabled`, though it would want a changelog note for self-hosters.

## Also

Both variables are now in `apps/web/.env.example`. #1278 notes the two gating lines were their only references anywhere in the repo — no example entry, no docs.

## Verification

- `biome check` passes on both changed files.
- `tsc --noEmit` on `apps/web`: 84 errors before, 84 after, **0 in the login page**. (Those 84 are pre-existing — CI only type-checks `@supermemory/ai-sdk` and `@supermemory/memory-graph`, which is the subject of #1249 / #1446.)

### DIFF
--- a/apps/web/.env.example
+++ b/apps/web/.env.example
@@ -2,3 +2,8 @@ NEXT_PUBLIC_BACKEND_URL=https://api.supermemory.ai
 NEXT_PUBLIC_POSTHOG_KEY=
 EXA_API_KEY=
 XAI_API_KEY=
+
+# Social sign-in providers. Both are shown by default; set a flag to "false" to
+# hide that provider's button on self-hosted deployments.
+NEXT_PUBLIC_GOOGLE_AUTH_ENABLED=
+NEXT_PUBLIC_GITHUB_AUTH_ENABLED=
--- a/apps/web/app/(auth)/login/page.tsx
+++ b/apps/web/app/(auth)/login/page.tsx
@@ -25,6 +25,25 @@ import {
 	shouldUseReviewerPasswordLogin,
 } from "@/lib/reviewer-password-login"
 
+/**
+ * `NEXT_PUBLIC_<PROVIDER>_AUTH_ENABLED` is an opt-out switch, not an opt-in one:
+ * self-hosted deployments show every social provider by default, and an operator
+ * turns one off by setting the flag to a falsy value. Treating any other value as
+ * "enabled" keeps `..._AUTH_ENABLED=true` from hiding the provider it names.
+ */
+function isSocialProviderEnabled(flag: string | undefined): boolean {
+	if (flag === undefined || flag === "") return true
+	return flag !== "false" && flag !== "0"
+}
+
+const isGoogleAuthEnabled =
+	process.env.NEXT_PUBLIC_HOST_ID === "supermemory" ||
+	isSocialProviderEnabled(process.env.NEXT_PUBLIC_GOOGLE_AUTH_ENABLED)
+
+const isGithubAuthEnabled =
+	process.env.NEXT_PUBLIC_HOST_ID === "supermemory" ||
+	isSocialProviderEnabled(process.env.NEXT_PUBLIC_GITHUB_AUTH_ENABLED)
+
 function isMcpOAuthAuthorizeContext(sp: Pick<URLSearchParams, "get">): boolean {
 	return sp.get("response_type") === "code" && Boolean(sp.get("client_id"))
 }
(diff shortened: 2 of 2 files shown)


Chirag6722:
@Dhravya @MaheshtheDev — first contribution here, so both Actions runs are parked at `action_required` and need a maintainer to approve them before anything runs.

In the meantime I ran CI's exact steps locally against this branch:

- `bunx biome ci --changed --since=main --no-errors-on-unmatched` — passes
- `bunx turbo run check-types --filter='@supermemory/ai-sdk' --filter='@supermemory/memory-graph'` — 2 successful, 2 total
- `bun.lock` is untouched, so `bun install --frozen-lockfile` is unaffected

The one thing worth your call is the default, which is why #1278 was filed as an issue rather than a PR: I went with **opt-out** semantics so no existing self-hosted deployment changes behaviour (unset still shows the provider, `true` now shows it instead of hiding it, `false`/`0` hides it). If you'd rather have true opt-in — unset means hidden — it's a two-line change to `isSocialProviderEnabled` and I'll push it right away.

Chirag6722:
Hi @MaheshtheDev — no problem with this being closed, but could you say what the reason was? There's no note here and I'd like to know for next time.

Asking because #1278 is still open and both gates are still inverted on `main` (`apps/web/app/(auth)/login/page.tsx` lines 479 and 536), so I don't think it was fixed elsewhere.

If it was the semantics: #1467 added the AgentID gate as plain opt-in (`HOST_ID === "supermemory" || NEXT_PUBLIC_AGENTID_AUTH_ENABLED`, no `!`), which answers the question #1278 was waiting on. I'd happily redo this to match that, so all three providers read the same way — it's a small change and the branch also needs a rebase now.

If you'd rather not take this from outside the team, that's completely fine — just let me know and I'll leave #1278 alone.

MaheshtheDev:
yeah, will share more briefing about it by monday

PR #1463, opened 2026-08-13
chore: make lint and format validation commands non-mutating

### LINKED ISSUE #1448: fix: make lint and format validation commands non-mutating
## Problem

The root `format-lint` script and the web app’s `lint` script run Biome with `--write`.

```json
"format-lint": "bunx biome check --write"
"lint": "biome check --write"
```

These commands are named and commonly used as validation tasks, but they silently modify files. This is unsafe in CI and surprising during routine local checks.

## Expected behavior

Validation commands should report formatting and lint problems without changing the working tree.

Auto-fixing should be an explicit opt-in command.

## Proposed fix

- remove `--write` from validation scripts:
  - root `format-lint`
  - `apps/web` `lint`
- add separate `:fix` scripts for intentional auto-fixes:
  - `format-lint:fix`
  - `lint:fix`

## Acceptance criteria

- validation commands leave `git status` unchanged
- validation commands exit non-zero when formatting or lint violations exist
- explicit `:fix` commands apply Biome fixes

Chirag6722:
Fixes #1448.

`format-lint` (root) and `lint` (`apps/web`) both ran `biome check --write`. CONTRIBUTING points at `bun run format-lint` as a pre-PR *validation* step, but running it rewrites your working tree — surprising locally, and unusable as a check in CI.

## Change

| script | before | after |
|---|---|---|
| `format-lint` | `biome check --write` | `biome check` |
| `format-lint:fix` | — | `biome check --write` |
| `apps/web` `lint` | `biome check --write` | `biome check` |
| `apps/web` `lint:fix` | — | `biome check --write` |

Docs updated to match in `CONTRIBUTING.md` and `CLAUDE.md`.

One non-obvious follow-on: `.github/workflows/claude-auto-fix-ci.yml` instructs the auto-fix bot to run `bun run format-lint` to *repair* lint failures. That only worked because of the `--write`, so it now points at `format-lint:fix` — otherwise this change would have silently broken the auto-fix job.

## Acceptance criteria

- **Validation commands leave `git status` unchanged** — verified for both `bun run format-lint` and `apps/web`'s `bun run lint`; tree byte-identical before and after.
- **Validation commands exit non-zero on violations** — both exit 1.
- **Explicit `:fix` commands apply fixes** — `format-lint:fix` / `lint:fix` retain `--write`.

## Note on the exit code

`bun run format-lint` exits 1 on current `main` — but so does `biome check --write` today, so this PR doesn't change that. There are 3 pre-existing errors, two of which are deserialization failures in the nested `apps/web/biome.json` and `packages/ui/biome.json`:

```
apps/web/biome.json:4:13     deserialize
packages/ui/biome.json:4:13  deserialize
```

Both pin `$schema` to `https://biomejs.dev/schemas/2.2.2/schema.json` while the workspace resolves `@biomejs/biome` at `^2.2.0` to a newer release. It doesn't surface in CI because `biome ci --changed` only ever inspects files touched by a PR. Left alone here to keep this PR to one concern — happy to open a separate one if you want it fixed.


suryaraj09:
## Review (manual verification)

Reviewed the diff against a fresh checkout of `main`. Thanks for splitting the docs/workflow updates out — that makes this the complete fix for #1448.

### Core change — verified 
- `biome check` without `--write` **is read-only**: ran it locally, file hashes unchanged after.
- It **exits non-zero** on violations, satisfying the issue's acceptance criteria.
- CI is unaffected: `.github/workflows/ci.yml` already uses `bunx biome ci --changed` (read-only).

### The workflow update is functionally necessary 
`claude-auto-fix-ci.yml` telling the auto-fix bot to run `format-lint:fix` isn't just docs polish — without it, the bot would run the now-read-only `format-lint` and be unable to apply any fix. Good catch including it.

### One gap worth a follow-up 
The current codebase has **652 pre-existing Biome errors** (mostly formatter output on `tsconfig.json` / `biome.json` files, ~73 parse errors, 2 deserialize errors). After this merges:
- `bun run format-lint` fails on a fresh `main` checkout (breaking the "run format-lint before submitting" flow in CONTRIBUTING.md)
- `format-lint:fix` would produce a ~600-file reformat diff

Root causes I found:
- `apps/web/biome.json` and `packages/ui/biome.json` pin `$schema` **2.2.2** while the repo runs Biome **2.4.6** → the deserialize diagnostics
- `tsconfig.json`/`biome.json` files were never Biome-formatted (indent mismatch)

Suggest a follow-up that either excludes config files from Biome formatting or formats-and-commits them once, so the documented `format-lint` gate is actually green on `main`.

Chirag6722:
Thanks for digging into this properly — and the `$schema` find is a good one. `apps/web/biome.json` and `packages/ui/biome.json` both pin `2.2.2` while the workspace resolves Biome to `2.4.6`, which is exactly where the two `deserialize` diagnostics come from. You're also right that `format-lint` is red on a fresh `main`, which is the part worth a follow-up.

On the count, though, I get a very different number. Measured on a pristine worktree of `main` with Biome 2.4.6:

```
Checked 651 files. Found 4 errors. Found 129 warnings.
```

and `biome check --write` on that same worktree fixes **exactly 1 file** (`apps/memory-graph-playground/next-env.d.ts`) — not a ~600-file reformat. No parse errors at all.

I think 652 is a line-endings artifact rather than a real property of the repo. On Windows with git's default `core.autocrlf=true`, every file is checked out CRLF and fails Biome's formatter on line endings alone, which lights up essentially the whole tree — including the `tsconfig.json` / `biome.json` "indent mismatch" ones. I hit this exact thing on this branch: 648 errors before I fixed it, 4 after. Worth re-running on your side with the tree normalized:

```bash
git -c core.autocrlf=input worktree add ../sm-clean origin/main --detach
bunx biome check ../sm-clean
```

So I'd frame the follow-up as narrower than you suggested: bump the pinned `$schema` in those two nested configs (and the one `next-env.d.ts`), and `format-lint` should be green on `main` — no repo-wide reformat and no need to exclude config files from the formatter. Happy to open that as a separate PR once this one lands, since it's a different concern.

Either way, this PR doesn't change the exit code on `main` — `biome check --write` exits 1 there today too, so nothing regresses by dropping `--write`.

PR #1473, opened 2026-08-14
fix(web): drop debug console.log that logged note content on every keystroke

abhay-codes07:
`FullscreenNoteModal`'s `handleContentChange` ran `console.log("handleContentChange", newContent)` on every change, so the full note body was written to the production console on every keystroke. Removed the leftover debug line; the surrounding `setContent` / `setDraft` behavior is unchanged.

Minor, but it is noisy and needlessly puts user note content in the console.

### DIFF
--- a/apps/web/components/fullscreen-note-modal.tsx
+++ b/apps/web/components/fullscreen-note-modal.tsx
@@ -61,7 +61,6 @@ export function FullscreenNoteModal({
 
 	const handleContentChange = useCallback(
 		(newContent: string) => {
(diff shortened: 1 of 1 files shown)


MaheshtheDev:
Included in #1596 with your commit and authorship intact, plus a `Co-Authored-By` trailer in the description so the credit survives the squash merge.

Good catch on this one. That line printed the whole note body on every keystroke, and PostHog session replay can pick console output up.

Thanks @abhay-codes07.

PR #1476, opened 2026-08-15
chore: remove unused modules

abhay-codes07:
Three modules have no remaining references anywhere in the repo (verified with `git grep` across all tracked source, excluding auto-generated WXT type stubs):

- **`packages/lib/glass-effect-manager.ts`** — a WebGL glass-effect singleton. Nothing imports it, and the live glass UI (`packages/ui/other/glass-effect.tsx`, used by `upgrade-mcp`) does not use it. (It also has a latent bug: the render loop is canceled once the last effect unmounts and `registerEffect` never restarts it — but since it is unused, deleting is cleaner than fixing.)
- **`useDeleteDocument`** in `packages/lib/queries.ts` — an unused delete-document mutation hook. Removing it also drops the now-unused `@tanstack/react-query`, `sonner`, `zod` and `$fetch` imports; the subscription-status helpers that remain are pure and need none of them.
- **`apps/browser-extension/utils/memory-popup.ts`** — a memory-popup helper referenced only by auto-generated WXT import stubs, never by a content script.

Net: **-498 lines**, no behavior change. Verified: `packages/lib`/`apps/web` type-check unaffected; `wxt prepare` regenerates clean (the stubs no longer mention memory-popup).

Happy to keep any of these if they are planned for upcoming work — just say and I'll drop it from the PR.


MaheshtheDev:

PR #1488, opened 2026-08-15
fix(tools): handle malformed OpenAI tool arguments

abhinav7x94:
## What changed

- return the existing structured error shape when OpenAI tool arguments contain malformed JSON
- keep one malformed call from rejecting an entire tool-call batch
- add focused offline regression coverage

## Why

OpenAI-compatible models can emit partial or invalid JSON in function arguments. The executor previously parsed outside its result flow, so a `SyntaxError` escaped and `Promise.all` discarded sibling results.

## Validation

- `bunx vitest run src/openai/tools.test.ts src/tool-operations.test.ts --testTimeout 100000` (16 passed)
- `bun run build`
- `bunx biome check packages/tools/src/openai/tools.ts packages/tools/src/openai/tools.test.ts`
- `git diff --check`

The package-wide TypeScript command remains red on unrelated pre-existing dependency and example errors; it reports no errors in the new test or changed lines.

### DIFF
--- a/packages/tools/src/openai/tools.test.ts
+++ b/packages/tools/src/openai/tools.test.ts
@@ -1,0 +1,58 @@
+import type OpenAI from "openai"
+import { describe, expect, it } from "vitest"
+import { createToolCallExecutor, createToolCallsExecutor } from "./tools"
+
+function toolCall(
+	name: string,
+	args: string,
+	id: string,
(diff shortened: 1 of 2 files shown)


MaheshtheDev:

We dropped `tools.test.ts`. The `parseToolArguments` change itself is unchanged.

Thanks @abhinav7x94.

PR #1498, opened 2026-08-16
fix(web): isolate query caches by auth scope

abhinav7x94:
## Summary

- recreate the app QueryClient whenever the authenticated session, user, organization, or restoration phase changes
- keep both the resolved organization and Better Auth's active organization in the scope so either side changing first closes the stale-cache window
- remount Autumn inside the same boundary so its private customer cache cannot survive an account or organization switch
- keep PostHog and error tracking outside the keyed remount
- add a mounted A → B → A regression and run the web unit suite in pull-request CI

This prevents fresh cached results under shared keys such as `sm_project_default` from being reused after a client-side auth or organization transition.

## Validation

- `cd apps/web && bun test` — 63 passed
- `cd apps/web && bun run build` — passed
- focused mounted/cache tests — 12 passed
- Biome and `git diff --check` — passed
- the existing web typecheck still reports unrelated baseline errors; no diagnostic references a changed file

Auth-scope transitions intentionally remount the data/app subtree. Analytics and global error tracking remain mounted.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -29,5 +29,9 @@ jobs:
       - name: Run TypeScript type checking
         run: bunx turbo run check-types --filter='@supermemory/ai-sdk' --filter='@supermemory/memory-graph'
 
+      - name: Run web unit tests
+        working-directory: apps/web
+        run: bun test
+
       - name: Run Biome CI (format & lint on changed files)
         run: bunx biome ci --changed --since=origin/main --no-errors-on-unmatched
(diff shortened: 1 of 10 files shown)


abhinav7x94:
Added a follow-up in [`a32811d`](https://github.com/abhinav7x94/supermemory/commit/a32811de744ffed648e4f5a10200dba3fd58d5be) after testing the OAuth consent flow against this cache boundary. Organization selection on `/oauth/consent` (including the trailing-slash route) now preserves the consent transaction, while session/account changes and organization changes everywhere else still create a fresh QueryClient.

Validation: 16/16 focused provider/config tests, 67/67 web tests, production Next build, Biome, and `git diff --check`; the existing app-wide typecheck has no diagnostics in changed files.

MaheshtheDev:

PR #1499, opened 2026-08-16
fix(web): scope organization state by session

abhinav7x94:
## Summary

- replace Better Auth's cross-session organization-list singleton with state keyed to the settled session and user
- synchronously hide prior-session organizations and active organization, rejecting late responses and enforcing latest-request-wins refreshes
- route OAuth consent through the scoped state and cancel stale tag or submit work during account and organization changes
- run the shared auth tests and package typecheck in pull-request CI

Better Auth 1.3.3 does not invalidate `useListOrganizations()` on sign-out/sign-in. In an SPA account A → logout → account B transition, B could receive A's cached organization list and the provider could attempt to restore A's active organization. OAuth consent also read that singleton directly.

This complements #1498 (TanStack Query/Autumn cache lifetime) and #1274 (persistent highlights Cache Storage); neither scopes Better Auth's organization atom.

## Validation

- `bun test packages/lib` — 11 passed
- `bun test apps/web` — 51 passed
- exact CI TypeScript filters — passed
- optimized Next.js production build — passed
- Biome and `git diff --check` — passed
- the existing app-wide typecheck baseline has no diagnostics in changed files

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -27,7 +27,10 @@ jobs:
         run: bun install --frozen-lockfile
 
       - name: Run TypeScript type checking
-        run: bunx turbo run check-types --filter='@supermemory/ai-sdk' --filter='@supermemory/memory-graph'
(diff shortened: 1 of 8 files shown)


MaheshtheDev:

PR #1507, opened 2026-08-16
fix(tools): await VoltAgent conversation persistence

abhinav7x94:
## Summary

- await VoltAgent conversation persistence before the `onEnd` lifecycle hook resolves
- preserve fail-open behavior because save failures are still caught and logged
- add deferred-fetch regressions for delayed success and delayed HTTP failure

## Why

The hook previously detached `saveConversation()`. In serverless or short-lived runtimes, the request could terminate after `onEnd` resolved but before `/v4/conversations` completed, silently dropping conversation memory.

## Validation

- `bunx vitest run src/voltagent/hooks.test.ts` (2/2)
- runnable tools suite: 92 passed, 30 skipped
- package build
- changed-file TypeScript check
- Biome and `git diff --check`

The two full-suite load failures are unchanged from main (missing `SUPERMEMORY_API_KEY` and the existing `./claude-memory` import).

### DIFF
--- a/packages/tools/src/voltagent/hooks.test.ts
+++ b/packages/tools/src/voltagent/hooks.test.ts
@@ -1,0 +1,113 @@
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest"
+import { createSupermemoryHooks } from "./hooks"
+import type { HookEndArgs } from "./types"
+
+const TEST_ARGS: HookEndArgs = {
+	agent: { name: "test-agent" },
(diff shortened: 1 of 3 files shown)


MaheshtheDev:

We dropped `hooks.test.ts`. Awaiting the save in `onEnd` is unchanged, and it was the most valuable of the batch since a torn-down serverless runtime was silently losing conversations.

Thanks @abhinav7x94.

PR #1514, opened 2026-08-16
fix(web): handle plugin key revocation errors

abhinav7x94:
## Summary

- Handle Better Auth API-key deletion errors returned in resolved responses.
- Route plugin disconnection through one guarded helper from both Integrations surfaces.
- Show success and refetch only after a confirmed successful revocation.
- Add regression coverage for resolved errors, rejected requests, fallback messages, success, and both production callers.
- Run the web unit suite in CI so the regression is merge-protected.

## Why

Better Auth 1.3.3 does not reject non-2xx responses by default. It resolves them as `{ data: null, error }`. Both plugin disconnect handlers ignored that result, so a failed credential revocation could leave the API key active while telling the user “Plugin disconnected.”

## Test coverage

- resolved delete error suppresses success and refetch
- rejected delete request preserves the existing failure path
- missing server error text uses the fallback message
- successful delete triggers success and one refetch
- both production handlers delegate through the guard

## Verification

- `bun test` from `apps/web` — 57 passed, 0 failed
- exact Bun 1.3.6 CI command — passed
- targeted strict TypeScript — passed
- changed-file Biome and `git diff --check` — passed
- full web typecheck remains baseline-red, with zero diagnostics in changed files

## Scope

This PR fixes false revocation success. Broader API-key list reconciliation and duplicate-click behavior are pre-existing follow-up concerns and are not changed here.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -29,5 +29,9 @@ jobs:
       - name: Run TypeScript type checking
         run: bunx turbo run check-types --filter='@supermemory/ai-sdk' --filter='@supermemory/memory-graph'
 
+      - name: Run web unit tests
+        working-directory: apps/web
+        run: bun test
+
       - name: Run Biome CI (format & lint on changed files)
         run: bunx biome ci --changed --since=origin/main --no-errors-on-unmatched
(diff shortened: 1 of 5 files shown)


MaheshtheDev:

PR #1525, opened 2026-08-17
fix(memory-graph): deduplicate shared graph entities

abhinav7x94:
## Summary

- materialize one graph node per memory ID even when a memory belongs to multiple source documents
- retain every document-to-memory derives edge and union relation edges across source occurrences
- normalize repeated documents across paginated responses so node counts, pagination, and rendering use the same identity model
- count unique document/memory IDs for `maxNodes` and pagination after container filtering
- use collision-free length-prefixed IDs for derives and relation edges
- run the memory-graph and web regression suites in CI

## Why

The API supports memories with multiple document sources. The graph previously treated each source occurrence as a separate node, so D3 linked to one duplicate while selection resolved another, leaving a disconnected phantom. Duplicate occurrences also consumed the node budget and could stop pagination before later unique nodes were fetched.

## Validation

- memory-graph: 201/201 tests
- web: 55/55 tests
- memory-graph typecheck and production build
- Next production build
- Biome and `git diff --check`
- regression-first coverage for shared IDs, repeated paginated documents, edge-ID collisions, cache/rerender stability, and under-budget pagination wiring

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -29,5 +29,13 @@ jobs:
       - name: Run TypeScript type checking
         run: bunx turbo run check-types --filter='@supermemory/ai-sdk' --filter='@supermemory/memory-graph'
 
+      - name: Run memory graph unit tests
+        working-directory: packages/memory-graph
+        run: bun run test
+
+      - name: Run web unit tests
+        working-directory: apps/web
+        run: bun test
+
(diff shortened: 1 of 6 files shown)


MaheshtheDev:
#1651 removed `apps/web`. Please drop the `apps/web` files from this PR and keep the rest. I'll review once that's done.

PR #1178, opened 2026-06-28
feat(web): company brain onboarding redesign + connector entitlement

MaheshtheDev:
- onboarding: unified About step, mode-aware Sources, Slack-focused Flows step, connect feedback (toast + connected state), auto-draft company description from domain
- brain-home: Active members stat + invite, real OneDrive icon
- useConnectorAccess hook so company_brain unlocks pro-tier connectors across onboarding + integrations
- fix company-brain-connections crash on empty connections

---
**Session Details**
- Session: [View Session](https://supermemory.us1.vorflux.com/agent-sessions/64a15b29-0848-42d1-af4c-138c5a27136f)
- Requested by: Unknown
- Address comments on this PR. Add `(aside)` to your comment to have me ignore it.

PR #1067, opened 2026-06-08
feat(web): make brain onboarding the default flow

MaheshtheDev:
feat(web): make brain onboarding the default flow

- Replace legacy /onboarding with the brain onboarding flow (about -> sources -> ingest -> team -> done)
- Add brain-home dashboard page and widgets
- Old /onboarding/brain deep links fall through to the existing catch-all redirect

feat(onboarding): create org on About step + remove skip

- Create the organization on About -> Continue (authClient.organization.create + setActiveOrg), persisting brain context to org metadata; fixes the no-org redirect loop
- Show a Creating… state on Continue while the org is created
- Remove Skip on the required About step (name + workspace)
- Add generateOrgSlug/generateUsername helpers to onboarding-brain/types

PR #1528, opened 2026-08-17
fix(web): bound and validate the onboarding extract-content batch

SEPURI-SAI-KRISHNA:
## Summary

`/api/onboarding/extract-content` forwards a caller-supplied `urls` array straight to
`https://api.exa.ai/contents` with `livecrawl: "fallback"`. The only validation today is
that every element is a non-empty string:

```ts
if (!urls.every((url) => typeof url === "string" && url.trim())) { ... }

body: JSON.stringify({ urls, text: true, livecrawl: "fallback" })
```

The array is unbounded, entries are never parsed or de-duplicated, and the scheme is
never checked. Each entry is a potentially live-crawled — and billed — Exa request, so a
single call carrying thousands of URLs (or the same URL thousands of times) turns into a
proportionally large invoice.

## Changes

- Cap the batch at `MAX_URLS = 20` and reject larger payloads with `400`.
- Parse each entry with `new URL()` and reject malformed values with `400`, instead of
  passing junk through to a paid API.
- Require `http:` or `https:`, so schemes like `file:` / `data:` never reach Exa.
- De-duplicate on the normalized `href`, so a repeated URL is billed once.

Behaviour for well-formed input is unchanged, apart from the URLs being sent in
normalized form.

## Note for reviewers: this route appears to be unused

While checking the call site to pick a safe `MAX_URLS`, I could not find any caller for
this route in the monorepo. The only `/api/*` route the web client fetches is `/api/og`
(`apps/web/components/memories-grid.tsx:157`).

The same appears to hold for its two siblings, `/api/onboarding/research` and
`/api/onboarding/account-status`. All three were added in #672 (Jan 2026), and the
onboarding flow has since been rebuilt more than once — #904 (`remove unused old
onboarding flow`), #1067, #1178 — which looks like it dropped the callers and left the
routes in place.

I have only grepped this repository, so I cannot rule out a caller outside the monorepo,
which is why this PR hardens rather than removes. If these routes are in fact dead,
deleting all three would be the better fix — it removes a network-reachable surface that
spends money against Exa and xAI — and I am happy to open that PR instead.

## Testing

`biome check` passes on the changed file. `apps/web` has no test runner configured (no
`vitest` dependency, no `test` script), so no tests were added.

### DIFF
--- a/apps/web/app/api/onboarding/extract-content/route.ts
+++ b/apps/web/app/api/onboarding/extract-content/route.ts
@@ -16,6 +16,10 @@ if (!exaApiKey) {
 	)
 }
 
+// Each URL becomes a (potentially billed live-crawled) Exa request, so cap the
+// batch size to bound cost/abuse from a single call.
+const MAX_URLS = 20
+
 export async function POST(request: Request) {
 	try {
 		if (!exaApiKey) {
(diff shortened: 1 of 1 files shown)


MaheshtheDev:
Thanks @SEPURI-SAI-KRISHNA — closing this in favour of #1589, which bundles the extract-content bounds together with the session-verification fix for the same route. That route was reachable unauthenticated, so the two changes needed to land together.

Your URL handling was better than what we had: #1589 now parses with `new URL()`, checks `protocol`, and de-duplicates before calling Exa — adapted from this PR, with you credited as co-author on that commit. One difference: #1589 caps the batch at 10 rather than 20, and also caps each URL at 2048 characters.

#1589 is still in draft while it goes through review — happy to reopen this if that changes.

SEPURI-SAI-KRISHNA:
Thanks for picking this up in #1589, and for the note on the URL handling.

One small thing: the merged commit (`3b0fc9c9`) doesn't carry a `Co-authored-by:` trailer, so the credit lives only in the commit body and PR description and GitHub doesn't pick it up as attribution.

Not worth rewriting an already-merged commit over, so nothing to action here. Just flagging it for future cherry-picks.

Issue #1544, opened 2026-08-19
`bun run check-types` has never passed on main — `@repo/web` reports ~90 errors


rajarshidattapy:
**Severity:** Medium (the CI gate added in #1447 is red at HEAD)
**Files:** `apps/web/**`, `apps/web/tsconfig.json`

## Reproduction

```
$ bun install
$ cd apps/web && bunx tsc --noEmit
... 90 errors
```

## Description

`#1447` ("fix: add type checks for TypeScript workspaces") wired `check-types`
into every workspace, but the task fails at HEAD in three packages. This issue
covers `@repo/web`; see #1542 (`@repo/ui`) and #1545
(`@supermemory/tools`) for the other two.

Error breakdown by file (top offenders):

```
27  components/orbit-memory.tsx
15  ../../packages/memory-graph/src/hooks/use-graph-data.ts
10  components/chat/index.tsx
 9  stores/chat.ts
 5  components/app-experience.tsx
 4  components/memory-graph/graph-card.tsx
 3  components/user-profile-menu.tsx
 ...
```

Two distinct root causes:

**1. `noUncheckedIndexedAccess` mismatch across package boundaries (most errors).**
`apps/web/tsconfig.json` extends `@total-typescript/tsconfig/bundler/dom/app`,
which turns on `noUncheckedIndexedAccess`. `packages/memory-graph/tsconfig.json`
only sets `"strict": true`. `apps/web` compiles the memory-graph *source*
directly, so `packages/memory-graph/src/hooks/use-graph-data.ts` type-checks
clean in its own package and fails inside web:

```
../../packages/memory-graph/src/hooks/use-graph-data.ts(518,46): error TS18048: 'doc' is possibly 'undefined'
```

These are strictness artifacts, not live crashes (the index is bounded by
`docCount`), but they make the gate unusable.

**2. Genuine defects.** See #1541 (`res.key` on a better-auth result) and
#1548 (`@repo/lib/constants` does not resolve) — both were surfaced by
this run and are real.

## Suggested fix

Align the shared-package tsconfigs with the app's strictness (or stop compiling
package sources across the boundary), fix the two genuine defects, and only then
put `check-types` in required CI so it stays green.

MaheshtheDev:
Closing this. The code it refers to was removed in #1651. Thanks for reporting.

PR #1554, opened 2026-08-19
fix(web): unwrap the iOS Shortcut API key from the better-auth result

### LINKED ISSUE #1541: iOS Shortcut API key creation reads `res.key` instead of `res.data.key`, always yields `undefined`
**Severity:** High (feature is broken)
**File:** `apps/web/components/settings/integrations.tsx:147-160`

## Description

`authClient.apiKey.create()` (better-auth) resolves to `{ data, error }`, not the
key object directly. `createApiKeyMutation` unwraps it wrong:

```ts
const res = await authClient.apiKey.create({ ... })
return res.key            // always undefined
```

`onSuccess` then does `setApiKey(undefined)`, opens the modal, and calls
`handleCopyApiKey(undefined)`.

Every other call site in the repo unwraps it correctly:

| File | Handling |
|---|---|
| `apps/web/components/integrations/raycast-detail.tsx:60` | `res.data?.key` ✅ |
| `apps/web/components/integrations/shortcuts-detail.tsx:88` | `res.data?.key` ✅ |
| `apps/web/components/settings/api-keys.tsx:196` | `extractCreatedKey(res)` ✅ |
| `apps/web/components/settings/integrations.tsx:186` (Raycast, same file) | `res.data?.key` ✅ |
| **`apps/web/components/settings/integrations.tsx:155` (iOS Shortcut)** | **`res.key` ❌** |

The mutation still "succeeds", so the user gets a modal with an empty key and a
clipboard copy that silently no-ops — no error surfaces. A key *is* created
server-side each time, so retrying leaks orphaned keys into the org.

TypeScript already flags this:

```
components/settings/integrations.tsx(155,15): error TS2339:
  Property 'key' does not exist on type 'Data<{ key: string; ... }> | Error$1<...>'
```

## Suggested fix

Match the Raycast mutation 20 lines below it:

```ts
if (!res.data?.key) throw new Error("API key missing from response")
return res.data.key
```

rajarshidattapy:
Fixes #1541.

## Why?

`authClient.apiKey.create()` resolves to `{ data, error }`, not the key object directly. The iOS Shortcut mutation in `components/settings/integrations.tsx` unwrapped it wrong:

```ts
const res = await authClient.apiKey.create({ ... })
return res.key            // always undefined
```

Because `undefined` is a perfectly good resolved value, the mutation still *succeeded*. `onSuccess` then ran `setApiKey(undefined)`, opened the modal, and called `handleCopyApiKey(undefined)` — so the user gets an empty key field and a clipboard copy that silently no-ops, with no error anywhere. A server-side failure was invisible for the same reason: nothing threw, so `onError` never fired.

A key *is* created server-side on every attempt, so retrying leaves orphaned keys on the org.

This was the last remaining call site with the bug — the other four already unwrap correctly, including the Raycast mutation 20 lines below it in the same file:

| Call site | Handling |
|---|---|
| `integrations/raycast-detail.tsx:61` | `res.data.key` |
| `integrations/shortcuts-detail.tsx:89` | `res.data.key` |
| `settings/api-keys.tsx:196` | `extractCreatedKey(res)` |
| `settings/integrations.tsx:187` (Raycast) | `res.data.key` |
| **`settings/integrations.tsx:155` (iOS Shortcut)** | **`res.key`** |

## What?

Adopt the shape the three sibling mutations already use — surface `res.error`, guard the missing key, return `res.data.key`:

```ts
if (res.error)
    throw new Error(res.error.message ?? "Failed to create API key")
if (!res.data?.key) throw new Error("API key missing from response")
return res.data.key
```

Also guard `org?.id` up front, matching that same Raycast mutation. The prefix template interpolated an undefined org id into the literal string `"sm_undefined_"` rather than failing — a second, quieter defect in the same six lines.

No behavioural change to the Raycast path, the modal, or the copy handler.

## Testing

`apps/web` has no test runner wired up, so I verified the unwrap out-of-band rather than adding vitest and a config to the app for one mutation — happy to add the harness if you'd prefer it pinned in CI.

Old vs. new against a `{ data, error }` result:

```
ok   old: success -> key      returned undefined            <-- modal gets no key
ok   new: success -> key      returned "sm_org_abc_livekey"
ok   old: failure -> silent   returned undefined            <-- onError never fires
ok   new: failure -> throws   threw "quota exceeded"
```

`tsc --noEmit` on `apps/web` goes from **84 errors to 83** — the one removed is exactly this line:

```
components/settings/integrations.tsx(155,15): error TS2339:
  Property 'key' does not exist on type 'Data<{ key: string; ... }> | Error$1<...>'
```

No new errors. (The remaining 83 are pre-existing and unrelated — see #1544.)

### DIFF
--- a/apps/web/components/settings/integrations.tsx
+++ b/apps/web/components/settings/integrations.tsx
@@ -144,15 +144,22 @@ export default function Integrations() {
 
 	const createApiKeyMutation = useMutation({
 		mutationFn: async () => {
+			if (!org?.id) {
+				throw new Error("Organization ID is required")
+			}
+
 			const res = await authClient.apiKey.create({
 				metadata: {
-					organizationId: org?.id,
+					organizationId: org.id,
 					type: "ios-shortcut",
 				},
 				name: `ios-${generateId().slice(0, 8)}`,
-				prefix: `sm_${org?.id}_`,
+				prefix: `sm_${org.id}_`,
 			})
-			return res.key
+			if (res.error)
+				throw new Error(res.error.message ?? "Failed to create API key")
+			if (!res.data?.key) throw new Error("API key missing from response")
+			return res.data.key
 		},
 		onSuccess: (key) => {
 			setApiKey(key)


ishaanxgupta:
I cannot reproduce in production, probably this comes from dead code, closing it for now!

PR #1563, opened 2026-08-19
remove unused onboarding research endpoints

ishaanxgupta:
## What changed

- removed the unused `/api/onboarding/research` route
- removed the unused `/api/onboarding/extract-content` route


MaheshtheDev:
Included in #1596 with your commits and authorship intact, plus a `Co-Authored-By` trailer in the description so the credit survives the squash merge.

You were right that these were unreachable. We had just added session checks and input caps to both routes in #1589, so this removes code we had only guarded because it was exposed. We also dropped `EXA_API_KEY`, `XAI_API_KEY` and the `@ai-sdk/xai` dependency, since nothing reads them once the routes are gone.

Thanks @ishaanxgupta.

Issue #1549, opened 2026-08-19
`forgetMemoryRequest` drops its 30s timeout whenever a caller passes a signal


rajarshidattapy:
**Severity:** Low (latent — defeats the fix from #1451)
**File:** `packages/tools/src/shared/forget-memory.ts:35`

## Description

`#1451` ("fix(tools): bound memory forget requests") added a 30-second abort so
a hung `DELETE /v4/memories` can't wedge the tool call. The two signals are
combined with `??`, so they're mutually exclusive:

```ts
signal: options?.signal ?? AbortSignal.timeout(FETCH_TIMEOUT_MS),
```

Pass a cancellation signal and the request becomes unbounded again — exactly the
condition the commit set out to remove. A caller who wants *both* cancellation
and a timeout has no way to express it through this API.

No production call site passes `options` today (`ai-sdk.ts:332` and
`openai/tools.ts:490` both omit it), so this is latent rather than live. It
becomes a real hang the first time someone wires up cancellation.

## Suggested fix

Compose the two instead of choosing between them:

```ts
signal: options?.signal
    ? AbortSignal.any([options.signal, AbortSignal.timeout(FETCH_TIMEOUT_MS)])
    : AbortSignal.timeout(FETCH_TIMEOUT_MS),
```

`AbortSignal.any` is available in Node 20+ (the repo's `engines` floor), Bun,
and workerd. `packages/tools/src/tool-operations.test.ts` already exercises the
signal path, so a case asserting the timeout still fires with a caller signal
present is a one-line addition.

Adityakk9031:
Opened #1675 composing caller signals with `AbortSignal.timeout` via `AbortSignal.any` (also fixed the same pattern in MCP client's `getDocuments`).

samvallad33:
forgetMemoryRequest dropping its 30s timeout whenever a caller passes a signal lets forget hang indefinitely and can leave memory in limbo. Operators then cannot tell whether the forget completed or stalled. Would you keep the timeout layered with the caller signal so forget cannot run unbounded?

PR #1564, opened 2026-08-19
test(mcp): restore SupermemoryClient coverage and fix two blank error messages

### LINKED ISSUE #1550: MCP `SupermemoryClient` has no test file — #1406 shipped error-handling logic with tests that don't exist
**Severity:** Low (test-coverage gap)
**File:** `apps/mcp/src/server/client/index.ts` (~600 lines, no test)

## Description

`src/server/client/index.test.ts` was **deleted** in `a99cf4f7`
("fix mcp graph and file uploads", #1397, 2026-08-04):

```
$ git log --oneline --all -- apps/mcp/src/server/client/index.test.ts
a99cf4f7 fix mcp graph and file uploads (#1397)      <- deleted here
19e8f06c MCP Revamp (#1120) (#1380)                  <- added here
```

Ten days later `5ecbc263` (#1406, "surface real API error messages") states in
its PR body:

> Added tests: a 403 with a JSON error body surfaces the API's message; an
> empty-body 403 gets the scope-aware fallback.
> `vitest run src/server/client/index.test.ts` — 3 passed.

But `git show --stat 5ecbc263` lists exactly one file:
`apps/mcp/src/server/client/index.ts`. The test file was never restored, so the
claimed run was against a file that no longer exists on the branch.

Current state of the MCP unit suite:

```
$ cd apps/mcp && bunx vitest run src
✓ src/server/auth/index.test.ts   (10 tests)
✓ src/server/analytics.test.ts    (4 tests)
✓ src/server/space.test.ts        (4 tests)
✓ src/server/auth/rbac.test.ts    (3 tests)
Test Files  4 passed (4)
```

`SupermemoryClient` — which owns every outbound Supermemory API call, the
`extractApiErrorMessage` unwrapping, and all the status-aware `handleError`
fallbacks — has zero coverage.

## Suggested fix

Restore a `src/server/client/index.test.ts` covering at minimum the two cases
#1406 claimed: a 403 with `{"error": "..."}` surfacing the API's message, and an
empty-body 403 falling back to the scope-aware text. `19e8f06c` has the original
file if it's a useful starting point.

addyCooks:
Closes #1550

`src/server/client/index.test.ts` was deleted in a99cf4f (#1397). Ten days
later #1406 added the API-error unwrapping and the status-aware `handleError`
fallbacks and claimed tests for them, but the file was never restored so
`SupermemoryClient`, which owns every outbound Supermemory API call, had no
coverage at all.

### Tests restored

53 cases in `src/server/client/index.test.ts`, with the `supermemory` SDK and
`fetch` both stubbed:

- SDK construction API key, base URL, 30s timeout, `x-sm-source` header
- Space scoping unscoped omits the filter, configured space is sent,
  explicit override wins, empty string is treated as unscoped
- Result normalisation `memory` / `chunk` / `context` / `content` precedence,
  chunk results stay distinguishable, 200k-char truncation
- `forgetMemory` exact match, 404 → similarity-search fallback, no match,
  chunks-only, non-404 errors do not fall back
- Raw-fetch endpoints request bodies, auth headers, caller-supplied abort
  signal, zod contract violations
- Full status table, timeouts, network failures, non-`Error` throws, and the
  `"<operation> failed: …"` prefixes

Includes the two cases #1406 described: a 403 with a JSON error body surfaces
the API's message, and an empty-body 403 gets the scope-aware fallback.

### Two fixes the tests forced (+2 lines in `index.ts`)

- **A 403 with body `{"error": ""}` leaked raw JSON to the user.**
  `extractApiErrorMessage` returns the raw string when the recognised key holds
  an empty value, so the user saw `{"error":""}`. A parsed envelope carrying no
  message now returns `undefined` and the caller reaches its fallback.
- **An unmapped status with an empty body produced a blank error message.**
  409/413/etc. fall past the `switch`, and `getDocuments`/`listMemoryEntries`
  construct `new Error("")` when the body is empty, so the user got nothing.
  Unmapped statuses with no message now report the status.

### Verification

- `vitest run src` 5 files, 74 tests passed (was 4 files, 21 tests)
- `tsc --noEmit -p tsconfig.json` clean, Biome clean
- Mutation-checked: 7 injected regressions each fail 1–4 of the new tests

### DIFF
--- a/apps/mcp/src/server/client/index.test.ts
+++ b/apps/mcp/src/server/client/index.test.ts
@@ -1,0 +1,668 @@
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest"
+import { DEFAULT_PROJECT_ID, getMemoryText, SupermemoryClient } from "./index"
+
+const { sdk, sdkOptions } = vi.hoisted(() => ({
+	sdk: {
+		add: vi.fn(),
+		profile: vi.fn(),
+		memories: { forget: vi.fn() },
+		search: { memories: vi.fn() },
+		documents: { list: vi.fn(), get: vi.fn() },
+	},
+	sdkOptions: vi.fn(),
+}))
+
+vi.mock("supermemory", () => ({
+	default: class {
+		constructor(options: unknown) {
+			sdkOptions(options)
+			Object.assign(this, sdk)
+		}
+	},
+}))
+
+const API_URL = "https://api.example.com"
+const TOKEN = "sm_test_token"
+const FORBIDDEN_FALLBACK = "read-only or scoped to specific spaces"
+
+const pagination = {
+	currentPage: 1,
+	limit: 50,
+	totalItems: 1,
+	totalPages: 1,
+}
+
+const documentsPayload = {
+	documents: [
+		{
+			id: "doc_1",
+			title: "Notes",
+			summary: null,
+			type: "text",
+			createdAt: "2026-01-01T00:00:00.000Z",
+			updatedAt: "2026-01-02T00:00:00.000Z",
+			memoryEntries: [],
+		},
+	],
+	pagination,
+}
+
+function client(containerTag?: string) {
+	return new SupermemoryClient(TOKEN, containerTag, API_URL)
+}
+
+function apiError(message: string, status: number) {
+	return Object.assign(new Error(message), { status })
+}
+
+describe("SupermemoryClient", () => {
+	let fetchMock = vi.fn()
+
+	beforeEach(() => {
+		fetchMock = vi.fn()
+		vi.stubGlobal("fetch", fetchMock)
+	})
+
+	afterEach(() => {
+		vi.unstubAllGlobals()
+		vi.clearAllMocks()
+	})
+
+	describe("configuration", () => {
+		it("identifies itself to the API with a source header and a timeout", () => {
+			client("work")
+
+			expect(sdkOptions).toHaveBeenCalledWith({
+				apiKey: TOKEN,
+				baseURL: API_URL,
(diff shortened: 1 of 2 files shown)


Agnik47:
Nice to see coverage here — #1550 has been open a while.

One thing worth catching before this lands: `uses a caller supplied abort signal and pagination` asserts `expect(init.signal).toBe(controller.signal)`, which pins the current behaviour at `apps/mcp/src/server/client/index.ts:339`:

```ts
const signal = options?.signal ?? AbortSignal.timeout(FETCH_TIMEOUT_MS)
```

That `??` makes cancellation and the 30s bound mutually exclusive — passing a signal removes the timeout, so a hung request is unbounded again.

It's the same defect as #1549 in `packages/tools`, fixed in #1595 by composing the two instead of choosing between them:

```ts
signal: options?.signal
```

No production caller in `apps/mcp` passes `options` today (`fetch-graph-data.ts` and `memory-graph.ts` both omit it), so it's latent either way — but a test asserting `toBe(controller.signal)` would make it meaningfully harder to fix afterwards, since the fix would then have to argue against a passing test.

Not asking you to expand scope here. Happy to send the `index.ts:339` fix as a follow-up once this merges — flagging it mainly so the assertion doesn't lock the behaviour in.

addyCooks:
Hey @Agnik47!
Good catch, fixed. 
Swapped the identity check for a behavioural one: 
it aborts the controller and asserts the signal `fetch` got observes it, 
so it passes now *and* after `AbortSignal.any`.

Sanity-checked both directions the old `toBe` fails under your fix, the new one doesn't. 
74/74 green either way. 
`index.ts:339` is all yours.

PR #1565, opened 2026-08-19
fix(tools): keep the forget timeout when a caller passes a signal

### LINKED ISSUE #1549: `forgetMemoryRequest` drops its 30s timeout whenever a caller passes a signal

rajarshidattapy:
Fixes #1549

## Problem

`forgetMemoryRequest` picked between the caller's signal and the 30s abort with `??`:

```ts
```

The two are mutually exclusive that way, so passing a cancellation signal silently drops the timeout and a hung `DELETE /v4/memories` can wedge the tool call again — the condition #1451 added the timeout to remove. There was also no way for a caller to ask for both.

Latent today: no production call site passes `options` (`ai-sdk.ts:332` and `openai/tools.ts:490` both omit it). It becomes a real hang the first time cancellation gets wired up.

## Fix

Compose the signals with `AbortSignal.any` instead of choosing between them, so cancellation and the timeout are both live. Available in Node 20+ (the repo's `engines` floor), Bun, and workerd.

## Test

Updated the existing signal case in `packages/tools/src/tool-operations.test.ts` to assert the composed behaviour: the signal handed to `fetch` is no longer the caller's (so the timeout is still attached), and aborting the caller's controller still aborts the request. All 14 tests in the file pass.

## Note (not in this PR)

`apps/mcp/src/server/client/index.ts:369` (`getDocuments`) has the identical `??` pattern. No caller there passes `options` either, so it's the same latent bug in a separate package — left out to keep this diff scoped to the issue; happy to fold it in if you'd prefer one PR.

### DIFF
--- a/packages/tools/src/shared/forget-memory.ts
+++ b/packages/tools/src/shared/forget-memory.ts
@@ -33,7 +33,12 @@ export async function forgetMemoryRequest(
 			Authorization: `Bearer ${apiKey}`,
 		},
 		body: JSON.stringify(params),
-		signal: options?.signal ?? AbortSignal.timeout(FETCH_TIMEOUT_MS),
+		signal: options?.signal
+			? AbortSignal.any([
+					options.signal,
+					AbortSignal.timeout(FETCH_TIMEOUT_MS),
+				])
+			: AbortSignal.timeout(FETCH_TIMEOUT_MS),
 	})
 
 	if (!response.ok) {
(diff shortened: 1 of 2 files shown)


MaheshtheDev:
Closing this one. Three PRs landed for the same change and we decided against it: `options?.signal ?? AbortSignal.timeout(...)` already honours a caller signal, and no production caller passes `options` today, so nothing can reach the path.

Thanks @rajarshidattapy.

PR #1566, opened 2026-08-19
fix(tools): stop memory file paths from colliding on one customId

### LINKED ISSUE #1547: Claude memory tool: path → customId normalization collides, silently overwriting unrelated memory files
**Severity:** Medium (silent data loss)
**File:** `packages/tools/src/claude-memory.ts:52-58`

## Description

`normalizePathToCustomId` flattens both `/` and `.` to `_`:

```ts
private normalizePathToCustomId(path: string): string {
    return path
        .replace(/^\//, "")   // strip leading slash
        .replace(/\//g, "_")  // / -> _
        .replace(/\./g, "_")  // . -> _
}
```

The mapping is not injective. All of these produce the identical customId
`memories_notes_txt`:

| Path | customId |
|---|---|
| `/memories/notes.txt` | `memories_notes_txt` |
| `/memories/notes_txt` | `memories_notes_txt` |
| `/memories/notes/txt` | `memories_notes_txt` |

And `/memories/project/a.md` collides with `/memories/project_a.md`.

Since `customId` is the document identity for `client.add()`, `create`,
`str_replace`, `insert`, `rename`, and `delete`, a collision means:

- **`create`** on the second path overwrites the first file's content with no
  warning — the tool reports success.
- **`view`** on either path returns whichever document won last.
- **`delete`** on either path destroys both.

Nested memory directories are the realistic trigger: Claude's memory tool
routinely creates paths like `/memories/projects/foo.md`, and any sibling
`/memories/projects_foo.md` (or a second nesting level that flattens the same
way) silently clobbers it.

## Suggested fix

Make the encoding reversible — escape the separator before substituting it, e.g.
`path.replace(/_/g, "__").replace(/[/.]/g, "_")`, or drop the transformation
entirely and hash the path (`customId = "mem_" + sha256(path).slice(0, 32)`) if
the backend accepts arbitrary customIds. Either way, add a round-trip assertion
for the three colliding paths above.

rajarshidattapy:
Fixes #1547

## Problem

`normalizePathToCustomId` flattened `/`, `.` and `_` all to `_`, so the mapping was not injective:

| Path | customId |
|---|---|
| `/memories/notes.txt` | `memories_notes_txt` |
| `/memories/notes_txt` | `memories_notes_txt` |
| `/memories/notes/txt` | `memories_notes_txt` |

`customId` is the document identity for `create`, `view`, `str_replace`, `insert`, `rename` and `delete`, so two paths sharing one id meant `create` silently overwrote an unrelated file, `view` returned whichever document won last, and `delete` destroyed both. Nested memory directories (`/memories/projects/foo.md` vs `/memories/projects_foo.md`) are the realistic trigger.

## Fix

Escape the separators instead of flattening them, making the encoding reversible:

```
__ = literal _    _s = /    _d = .
```

Every `_` in the output opens a two-character sequence, so decoding is unambiguous and distinct paths always get distinct ids. `/memories/notes.txt` now maps to `memories_snotes_dtxt`.

Hashing was the other option in the issue, but `getFileDocument` passes the customId as the semantic search query (`q`) and then exact-matches the result — a hex digest as the query would stop retrieving the document at all.

## Existing documents

Changing the encoding orphans anything written under the old ids, which would be its own silent data loss. So `getFileDocument` — the single lookup every command routes through — falls back to the legacy id, guarded by an exact `metadata.file_path` match, since legacy ids are the ambiguous ones and must not be trusted on their own. The next write to that path re-adds it under the new id, so the fallback self-heals rather than lingering. The search query stays the legacy (word-preserving) form, so search relevance is unchanged.

## Tests

Added to `packages/tools/src/claude-memory.test.ts`:
- the five colliding paths from the issue now produce five distinct customIds
- a legacy-id document with a matching `file_path` is still readable
- a legacy-id document with a *different* stored `file_path` is not served (reports not-found)

Existing tests that hardcoded the old ids were updated to the new encoding. `claude-memory.test.ts` and `tool-operations.test.ts`: 27/27 pass. The two failing files elsewhere in `packages/tools` (the stale `test/` directory importing `./claude-memory`) fail identically on `main` and are untouched here.

### DIFF
--- a/packages/tools/src/claude-memory.test.ts
+++ b/packages/tools/src/claude-memory.test.ts
@@ -24,9 +24,9 @@ const FILE_CONTENT = "line1\nline2\nline3\nline4\nline5"
 
 function mockDocument(content: string) {
 	// `readFile` matches by `documentId === normalizePathToCustomId(path)`.
-	// normalizePathToCustomId("/memories/notes.txt") -> "memories_notes_txt"
+	// normalizePathToCustomId("/memories/notes.txt") -> "memories_snotes_dtxt"
 	searchExecute.mockResolvedValue({
-		results: [{ documentId: "memories_notes_txt", content }],
+		results: [{ documentId: "memories_snotes_dtxt", content }],
 	})
 }
 
@@ -97,8 +97,8 @@ describe("ClaudeMemoryTool exact-file matching", () => {
 	it("view finds the exact file even when a neighbour ranks first", async () => {
 		searchExecute.mockResolvedValue({
 			results: [
-				{ documentId: "memories_notes_backup_txt", content: "backup stuff" },
-				{ documentId: "memories_notes_txt", content: FILE_CONTENT },
+				{ documentId: "memories_snotes__backup_dtxt", content: "backup stuff" },
+				{ documentId: "memories_snotes_dtxt", content: FILE_CONTENT },
 			],
 		})
 
@@ -117,7 +117,7 @@ describe("ClaudeMemoryTool exact-file matching", () => {
 		// be served as the requested one.
 		searchExecute.mockResolvedValue({
 			results: [
-				{ documentId: "memories_notes_backup_txt", content: "backup stuff" },
+				{ documentId: "memories_snotes__backup_dtxt", content: "backup stuff" },
 			],
 		})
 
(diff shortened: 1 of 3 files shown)


MaheshtheDev:
Closing as a duplicate of #1586, which covers the same customId collision. That one is still open while we work out a migration, since any change to the id scheme orphans memories already stored under the old one.

Thanks @rajarshidattapy.

Issue #1548, opened 2026-08-19
`@repo/lib/constants` does not resolve — three web components use an alias that isn't mapped


rajarshidattapy:
**Severity:** Low
**Files:** `apps/web/components/integrations/chrome-detail.tsx:3`,
`apps/web/components/nova/nova-empty-state.tsx:3`,
`apps/web/components/onboarding/x-bookmarks-detail-view.tsx:7`

## Description

Three files import the Chrome extension URL through an alias that neither
`apps/web/tsconfig.json` nor the `@repo/lib` exports map can resolve:

```ts
import { CHROME_EXTENSION_URL } from "@repo/lib/constants"
```

```
components/integrations/chrome-detail.tsx(3,38): error TS2307:
  Cannot find module '@repo/lib/constants' or its corresponding type declarations.
```

`apps/web/tsconfig.json` maps `@repo/ui/*`, `@ui/*`, `@lib/*`, and `@hooks/*` —
but not `@repo/lib/*`. Resolution then falls to the package's own exports map,
`{"./*": "./*"}`, which points at `packages/lib/constants` (extensionless); the
actual file is `constants.ts`, and TypeScript's `bundler` resolution does not
append extensions after an exports-map substitution.

Every other consumer in the repo uses the mapped alias:

```
apps/web/app/(app)/brain/page.tsx:10:       import { SHARED_TEAM_BRAIN_TAG } from "@lib/constants"
apps/web/components/app-experience.tsx:46:  import { DEFAULT_PROJECT_ID } from "@lib/constants"
apps/web/components/dashboard-view.tsx:23:  import { CHROME_EXTENSION_URL, RAYCAST_EXTENSION_URL } from "@lib/constants"
```

Webpack/Turbopack apply extension resolution more permissively than `tsc`, so
these likely still bundle at runtime — but the import is unsound and it is one
of the errors keeping `check-types` red (see #1544).

## Suggested fix

Change the three imports to `@lib/constants` to match the rest of the codebase.

MaheshtheDev:
Closing this. The code it refers to was removed in #1651. Thanks for reporting.

Issue #1542, opened 2026-08-19
`AnonymousAuth` is dead code that is broken at runtime and fails `check-types`


rajarshidattapy:
**Severity:** Medium (blocks `bun run check-types` for the whole monorepo)
**File:** `packages/ui/other/anonymous-auth.tsx`

## Description

The component treats every better-auth client call as if it returned raw data,
but the client returns `{ data, error }`. Every access is therefore `undefined`
at runtime:

```ts
const session = await authClient.getSession()
if (!session?.session) { ... }              // always true  -> always "no session"

const res = await authClient.signIn.anonymous()
if (!res.token) throw new Error(...)        // always throws -> catch -> router.push(loginPath)
```

So the component can only ever do one thing: throw and redirect to `/login`.
The org-listing branches (`orgs?.length`, `orgs?.map`, `orgs?.[0]`, `org.slug`)
are equally wrong — `authClient.organization.list()` returns `{ data, error }`
too.

`tsc` reports 21 errors in this one file, and because `@repo/ui` is a dependency
of `@repo/web` and `supermemory-mcp`, `bun run check-types` fails at the root:

```
@repo/ui:check-types: other/anonymous-auth.tsx(17,18): error TS2339: Property 'session' does not exist on type 'Data<...> | Error$1<...>'
... 20 more
ERROR  run failed: command exited (2)
Failed: @repo/ui#check-types
```

## Impact

`AnonymousAuth` has **zero importers** — `grep -rn "AnonymousAuth" apps packages`
matches only its own definition. It is unreferenced dead code that is
nonetheless red-lighting type checks for every consumer package.

## Suggested fix

Delete `packages/ui/other/anonymous-auth.tsx`. If anonymous sign-in is still
wanted later, it should be rewritten against the `{ data, error }` contract
rather than resurrected as-is.

PR #1571, opened 2026-08-19
fix(web): clear the 84 type errors blocking check-types

### LINKED ISSUE #1544: `bun run check-types` has never passed on main — `@repo/web` reports ~90 errors

rajarshidattapy:
Fixes #1544

## Result

`cd apps/web && bunx tsc --noEmit`: **84 → 3**, and the 3 that remain are the `@repo/lib/constants` alias errors from #1548, already fixed in #1569. With that branch applied on top I get **0** locally, so the two together turn the gate green for `@repo/web`.

`packages/memory-graph` still reports 0 and `packages/ui` still reports its pre-existing 23 (#1542) — neither changed.

## One expression accounted for a quarter of it

`useProject().selectedProject` was `normalizedProjects[0]`, so `string | undefined` — even though the line above guarantees the array is non-empty:

```ts
const normalizedProjects = selectedProjects.length === 0 ? [defaultTag] : selectedProjects
const selectedProject = normalizedProjects[0] ?? defaultTag   // was: normalizedProjects[0]
```

That one-line default cleared ~20 errors across `stores/chat`, `chat/index`, `add-document`, `app-experience`, `quick-note-card`, `fullscreen-note-modal` and `home-chat-composer`, all of which pass the project id into functions typed `string`.

## The rest

- **`orbit-memory.tsx` (27).** `ring` is now `0 | 1` and `RR`/`SPEED` are tuples, so ring lookups stay `number`; `INTEG` is iterated with `.entries()`; the comet/gradient refs are captured and skipped when absent rather than indexed blind; a node without a `key` renders an empty glyph instead of `undefined` HTML.
- **`use-graph-data.ts` (15) and `graph-card.tsx` (4).** `.entries()` for the document spiral, and edges whose endpoints are missing are skipped instead of pushing `undefined` coordinates.
- **Chat history buckets, user initials.** Optional access on fixed-shape arrays TypeScript can't narrow.
- **`stores/chat.ts`.** Removed `if (msgA.content !== msgB.content) return false` — `UIMessage` keeps its text in `parts`, so this compared `undefined` with `undefined` on every call and never fired; the `parts` comparison directly below is the real check. Also typed the conversations fallback so `Object.entries` yields `ConversationRecord` instead of `unknown`.
- **`useResetOrganization`.** `retry: { attempts: 0 }` → `retry: 0`; the object form of `RetryOptions` requires `type`/`baseDelay`/`maxDelay`, the number form is exactly "no retries".
- **`SyncLogoIcon`.** Accepts the `style` prop that `timeline-view.tsx` was already passing it.

## A real bug, not a strictness artifact

`components/settings/integrations.tsx` returned `res.key` from `authClient.apiKey.create()`, which resolves to `{ data, error }`. `res.key` was always `undefined`, so the iOS-shortcut flow set `undefined` into the key modal and handed `undefined` to `navigator.clipboard.writeText`. It now unwraps the result exactly as `createRaycastApiKeyMutation` twenty lines below already does. This is #1541.

## On the approach

The issue suggests aligning the shared-package tsconfigs with the app's strictness. I fixed the code instead: `packages/memory-graph/src/hooks/use-graph-data.ts` is compiled by `apps/web` from source, so making it index-safe is what actually removes the errors, and flipping `noUncheckedIndexedAccess` on in that package would only surface more of the same in files web doesn't compile. Loosening `apps/web` was the other option and would have buried the three genuine defects above.

Nothing here changes runtime behavior except the API-key fix. The unreachable `?? defaultTag`, the removed always-false comparison, and the `?.` guards on fixed-size arrays are all no-ops at runtime.

## Verification

`bun test` in `apps/web`: 51 pass / 0 fail. `biome ci --changed` exits 0.

### DIFF
--- a/apps/web/components/chat/index.tsx
+++ b/apps/web/components/chat/index.tsx
@@ -1612,11 +1612,11 @@ export function ChatSidebar({
 		]
 		for (const t of filtered) {
 			const ts = new Date(t.updatedAt).getTime()
-			if (ts >= startOfToday) buckets[0].items.push(t)
-			else if (ts >= startOfToday - day) buckets[1].items.push(t)
-			else if (ts >= startOfToday - 7 * day) buckets[2].items.push(t)
-			else if (ts >= startOfToday - 30 * day) buckets[3].items.push(t)
-			else buckets[4].items.push(t)
+			if (ts >= startOfToday) buckets[0]?.items.push(t)
+			else if (ts >= startOfToday - day) buckets[1]?.items.push(t)
+			else if (ts >= startOfToday - 7 * day) buckets[2]?.items.push(t)
+			else if (ts >= startOfToday - 30 * day) buckets[3]?.items.push(t)
+			else buckets[4]?.items.push(t)
 		}
 		return buckets.filter((b) => b.items.length > 0)
 	}, [threads, historySearch])
--- a/apps/web/components/memory-graph/graph-card.tsx
+++ b/apps/web/components/memory-graph/graph-card.tsx
@@ -67,15 +67,13 @@ export function StaticGraphPreview({
 		const result: { x1: number; y1: number; x2: number; y2: number }[] = []
 		const edgeCount = Math.min(nodes.length - 1, 20)
 		for (let i = 0; i < edgeCount; i++) {
-			const a = Math.floor(rand() * nodes.length)
-			let b = Math.floor(rand() * nodes.length)
-			if (b === a) b = (a + 1) % nodes.length
-			result.push({
-				x1: nodes[a]?.x,
-				y1: nodes[a]?.y,
-				x2: nodes[b]?.x,
-				y2: nodes[b]?.y,
-			})
+			const aIdx = Math.floor(rand() * nodes.length)
+			let bIdx = Math.floor(rand() * nodes.length)
+			if (bIdx === aIdx) bIdx = (aIdx + 1) % nodes.length
+			const a = nodes[aIdx]
+			const b = nodes[bIdx]
+			if (!a || !b) continue
+			result.push({ x1: a.x, y1: a.y, x2: b.x, y2: b.y })
 		}
 		return result
 	}, [nodes])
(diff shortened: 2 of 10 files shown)


MaheshtheDev:

PR #1572, opened 2026-08-20
Fix: document title fallback

### LINKED ISSUE #1425: Auto-titling regression: most MCP-connector saves land as "Untitled Document" since ~Aug 6
Since around August 6, most memories I save through the claude.ai MCP connector get the title "Untitled Document" instead of a generated title. Before that, titling was reliable: 14 of 14 saves on August 5 landed with clean titles taken from the content's first line.

Timeline from my account (Free plan, default project, saves via the connector's `add_memory`):

- Aug 5: 14/14 saves titled correctly (each opened with a short title line, then a blank line, then prose).
- Aug 6: 3 of 5 saves landed "Untitled Document" despite the same format. The two that titled got LLM-paraphrased titles, not the first line.
- Aug 7: controlled test, 3 of 3 failed. Two cards opened with a markdown H1 title, one with a YAML frontmatter `title:` block. All processed to `status: done` and landed "Untitled Document" (IDs `xmDzi7vS77T9N3rDxTANbA`, `XWY581gieqDPzqV6vUkaP1`, `sWZ5U2yYYtEnYRsPgookB2`, created 2026-08-07 ~12:53 UTC). All three shared the same `updatedAt` down to the millisecond, so the whole processing batch appears to have failed titling together.

So it doesn't look format-dependent; it looks like the titling step started failing silently. Content processing is otherwise fine (summaries and search work).

Two asks:

1. Is the titling regression known, and can it be fixed? Happy to share more IDs from my account for tracing.
2. Feature request: let callers set the title explicitly, either a `title` field on `POST /v3/documents` or honoring `metadata.title` in display. The API currently has no title control at all, so when auto-titling fails there's no way to repair a card short of deleting and re-saving.

addyCooks:
Fixes #1425

### What's happening

Auto-titling runs server-side during ingestion. When it fails, `documents.title`
stays null and every surface falls back to "Untitled Document" even though the
content itself opens with a perfectly good title. The reporter's three test cards
(markdown H1, YAML frontmatter, plain title line) all processed to `status: done`
with a null title.

This does not fix the server-side titling regression, which lives outside this
repo. It makes the null title recoverable, and gives callers the title control
the issue asks for.

### What this does

`resolveDocumentTitle()` resolves a display title in order:

1. `metadata.title` the caller's explicit choice, so it outranks auto-titling
2. the stored `title`
3. derived from content YAML frontmatter `title:`, a markdown heading, or a
   short standalone opening line

Derivation is deliberately conservative: list/quote/table openers, code fences,
horizontal rules, bare URLs, punctuation-only lines, and long opening paragraphs
all decline rather than produce a bad title.

Wired into the document modal, memory grid, MCP/note cards, dashboard recents,
brain-home recents, and the command palette.

### Title control

The ingest API has no `title` field, so `add_memory` now takes an optional
`title` and pins it through `metadata.title`, which the app and the MCP
`list_documents` / `get_document` output both read back. No API change needed 
this works against the API as it ships today.

A real `title` field on `POST /v3/documents` still needs a server-side change.

### Tests

- `apps/web/lib/document-title.test.ts` 22 cases covering precedence and the
  derivation edge cases (CRLF, BOM, setext headings, nested frontmatter keys,
  truncation, blank/non-string metadata)
- `apps/mcp/src/server/format.test.ts` 5 cases for pinned titles in MCP output
- `apps/mcp/src/server/client/index.test.ts` 2 cases for title pinning

### DIFF
--- a/apps/mcp/src/server/client/index.test.ts
+++ b/apps/mcp/src/server/client/index.test.ts
@@ -95,6 +95,34 @@ describe("SupermemoryClient", () => {
 			})
 		})
 
+		it("pins an explicit title through metadata", async () => {
+			sdk.add.mockResolvedValue({ id: "doc_2" })
+
+			await client("work").createMemory("remember this", {
+				title: "  Quarterly planning notes  ",
+			})
+
+			expect(sdk.add).toHaveBeenCalledWith({
+				content: "remember this",
+				containerTag: "work",
+				metadata: {
+					sm_source: "supermemory-mcp",
+					title: "Quarterly planning notes",
+				},
+			})
+		})
+
+		it("leaves the title out when it is blank or absent", async () => {
+			sdk.add.mockResolvedValue({ id: "doc_3" })
+
+			await client("work").createMemory("a", { title: "   " })
+			await client("work").createMemory("b", {})
+
+			for (const call of sdk.add.mock.calls) {
+				expect(call[0].metadata).toEqual({ sm_source: "supermemory-mcp" })
+			}
+		})
+
 		it("treats an empty space string as an unscoped connection", async () => {
 			sdk.search.memories.mockResolvedValue({
 				results: [],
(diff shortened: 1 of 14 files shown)


MaheshtheDev:
#1651 removed `apps/web`. Please drop the `apps/web` files from this PR and keep the rest. I'll review once that's done.

addyCooks:
Hey @MaheshtheDev, 
I’ve removed the apps/web files from the PR and kept the rest of the changes intact...
Feel free to review whenever you’re ready : )

PR #1587, opened 2026-08-22
fix(mcp): report transient session-endpoint failures as 503, not invalid token

### LINKED ISSUE #1551: MCP: a transient `/v3/session` failure is reported to the client as an invalid token
**Severity:** Low
**File:** `apps/mcp/src/server/auth/index.ts:70-93`

## Description

`validateApiKey` collapses every failure mode into `null`:

```ts
try {
    const session = await fetchSession(token, apiUrl)
    ...
} catch (error) {
    console.error("API key validation error:", error)
    return null
}
```

`fetchSession` throws on *any* non-OK status and on the 30s `AbortSignal.timeout`.
`handleMcpRequest` then maps `null` to `unauthorizedResponse(url, true)`:

```json
{"jsonrpc":"2.0","error":{"code":-32000,"message":"Invalid or expired token"},"id":null}
```
with `WWW-Authenticate: Bearer error="invalid_token"`.

So an API 500, a Hyperdrive blip, or a slow session endpoint tells the MCP client
its perfectly valid `sm_` key is invalid. Clients act on `invalid_token` by
discarding the credential and starting a re-auth flow — users get pushed through
the browser connect flow to fix an outage on our side.

The OAuth path has the same shape, but there the failure is local JWT
verification (no network), so `null` genuinely does mean "bad token". The API-key
path added in #1537 is the first one where `null` conflates "rejected" with
"couldn't ask".

`fetchSession` already attaches `status` to the thrown error, so the information
needed to distinguish the cases is present:

```ts
throw Object.assign(new Error(`Session request failed with status ${response.status}`),
    { status: response.status })
```

## Suggested fix

Distinguish 401/403 (genuinely invalid → `null` → `invalid_token`) from
everything else (transient → propagate, and have `handleMcpRequest` return 503
with `-32000 "Authentication service unavailable"`). Serving a stale cache entry
past its TTL during an outage would be a reasonable second layer, though it
extends the revocation window the PR already flagged.

Sravanjangam:
Fixes #1551.

Hi supermemory team 👋 Thanks for building such a great product! While running a security & quality audit of the repo we hit this issue and put together a small, tested fix — details below.

## Problem

`validateApiKey` collapsed **every** `fetchSession` failure — network errors, timeouts, any non-OK status — into `null`, which the transport maps to JSON-RPC `-32000 "Invalid or expired token"`. During an api.supermemory.ai/Hyperdrive outage this made MCP clients discard perfectly valid `sm_` keys and forced users through browser re-authentication at exactly the wrong moment, amplifying incidents into support load.

## Solution

Classify failures instead of collapsing them. A new `TransientAuthError` is thrown for timeouts (`AbortError`/`TimeoutError`) and upstream statuses other than 401/403; the transport maps it to a **503** JSON-RPC response with `Retry-After: 5`. Genuine rejections (401/403/schema failures/missing org) still resolve `null` → unchanged `invalid_token` behavior. Success-path caching untouched.

## Changes

- `apps/mcp/src/server/auth/index.ts` → `TransientAuthError` class + typed catch in `validateApiKey`
- `apps/mcp/src/server/index.ts` → try/catch around token validation → 503 response with `Retry-After`
- `apps/mcp/src/server/auth/index.test.ts` → +2 tests (500 → TransientAuthError; timeout → TransientAuthError)
- applied review feedback: auth resolution extracted into `resolveAuthUser()` returning a typed `AuthResolution` union — call site is all-`const`, behavior unchanged

## Verification

Fresh from the committed branch: `bunx vitest run src/server/auth` → **15/15 pass**, including the pre-existing revoked-key → `null` case proving rejection behavior is unchanged; `tsc` adds zero new errors vs baseline; Biome clean.

---
Happy to iterate on any of this — feedback and reworks very welcome! 🙏

## Environment

- macOS 26.1 (arm64) · bun 1.4.0 · node v26.7.0
- vitest 3.2.4 (workspace-pinned) · Biome lint clean
- Branch `fix/mcp-transient-session-errors` — all gates re-run fresh at commit `b561e12a27c5`

### DIFF
--- a/apps/mcp/src/server/auth/index.test.ts
+++ b/apps/mcp/src/server/auth/index.test.ts
@@ -1,6 +1,11 @@
 import { createLocalJWKSet, exportJWK, generateKeyPair, SignJWT } from "jose"
 import { afterEach, beforeAll, describe, expect, it, vi } from "vitest"
-import { fetchSession, validateApiKey, validateOAuthToken } from "./index"
+import {
+	fetchSession,
+	TransientAuthError,
+	validateApiKey,
+	validateOAuthToken,
+} from "./index"
 
 const API_URL = "https://api.example.com"
 const ISSUER = `${API_URL}/api/auth`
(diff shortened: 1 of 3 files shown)


MaheshtheDev:
Merged as part of #1591 with your commits and authorship intact. We extended it to the OAuth path too, since `validateOAuthToken` had the same swallow-everything catch and a JWKS failure still surfaced as `invalid_token`.

Thanks @Sravanjangam.

PR #1603, opened 2026-08-27
fix(tools): keep the forget-request timeout when a caller passes a signal

### LINKED ISSUE #1549: `forgetMemoryRequest` drops its 30s timeout whenever a caller passes a signal

Cintu07:
Fixes #1549

`forgetMemoryRequest` combined the caller's abort signal and the 30s fetch timeout with `??`, so the two were mutually exclusive. Passing a cancellation signal dropped the timeout and made the DELETE /v4/memories request unbounded again, undoing the bound #1451 added. A caller that wanted both cancellation and a timeout had no way to ask for it.

It is latent today, since no production call site passes options (ai-sdk.ts:332 and openai/tools.ts:490 both omit it). It becomes a real hang the first time someone wires up cancellation.

Changes:
- compose the caller signal and the timeout with AbortSignal.any instead of choosing between them, so the request aborts on whichever fires first
- fall back to the plain timeout when no caller signal is passed
- rewrite the existing signal test, which asserted the dropped-timeout behaviour, to assert the composed signal aborts on the caller and still carries the timeout

AbortSignal.any is available on the repo's Node 20 engines floor, Bun, and workerd.

Tested:
vitest run src/tool-operations.test.ts (14 passed)

### DIFF
--- a/packages/tools/src/shared/forget-memory.ts
+++ b/packages/tools/src/shared/forget-memory.ts
@@ -33,7 +33,12 @@ export async function forgetMemoryRequest(
 			Authorization: `Bearer ${apiKey}`,
 		},
 		body: JSON.stringify(params),
-		signal: options?.signal ?? AbortSignal.timeout(FETCH_TIMEOUT_MS),
+		// compose the caller's signal with the timeout instead of choosing
+		// between them: `??` dropped the 30s bound whenever a signal was passed,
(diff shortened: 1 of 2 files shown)


MaheshtheDev:
duplicate of #1595

PR #1608, opened 2026-08-27
fix(tools): make Claude memory tool exact-file lookup and listing deterministic

Adityakk9031:
This PR fixes the Claude memory tool file lookup and listing. It replaces semantic vector search with deterministic document list and get APIs, avoiding mismatch between internal UUID documentId and developer customId.


Adityakk9031:
@MaheshtheDev have a look

PR #1594, opened 2026-08-24
feat(tools): apiKey option, type re-exports, and two reliability fixes

MaheshtheDev:
Cherry-picks four contributor PRs for `@supermemory/tools` onto one branch, and bumps the package to 2.2.0.

- #1244 (@rajarshidattapy): `withSupermemory` accepts `options.apiKey` instead of only reading `SUPERMEMORY_API_KEY`, matching the Vercel, Mastra and Voltagent integrations. Unblocks secrets managers, edge runtimes and per-request keys.
- #1574 (@Agnik47): re-exports `PromptTemplate`, `MemoryPromptData` and `WithSupermemoryOptions` from `ai-sdk`. `./vercel` is not a published subpath, so the documented custom-template example did not compile.
- #1488 (@abhinav7x94): malformed tool-call JSON returns an error result instead of throwing out of the request.
- #1507 (@abhinav7x94): VoltAgent `onEnd` awaits the conversation save, which was fire-and-forget and could be dropped when a serverless runtime tore down.

Dropped the `middleware.test.ts` added by #1244. Note that editing `packages/tools/package.json` triggers the npm publish workflow on merge.

Co-Authored-By: rajarshidattapy <138959719+rajarshidattapy@users.noreply.github.com>
Co-Authored-By: Agnik47 <140933190+Agnik47@users.noreply.github.com>
Co-Authored-By: abhinav7x94 <204053250+abhinav7x94@users.noreply.github.com>

Issue #1545, opened 2026-08-19
`@supermemory/tools` fails `check-types`: 31 errors in `test/`, including an import of a file that no longer exists


rajarshidattapy:
**Severity:** Medium (blocks the monorepo type-check gate)
**Files:** `packages/tools/test/**`

## Reproduction

```
$ cd packages/tools && bunx tsc --noEmit
```

## Description

All 31 errors are in `test/`; `src/` is clean. Breakdown:

```
19  test/mastra/unit.test.ts
 5  test/test-memory-tool.ts
 5  test/mastra/integration.test.ts
 2  test/with-supermemory/unit.test.ts
```

Three separate causes:

**1. Stale Mastra fixtures (24 errors).** `ProcessInputArgs` /
`ProcessOutputResultArgs` gained a required `state` property in the installed
Mastra version; the hand-built fixtures never got it:

```
test/mastra/unit.test.ts(187,10): error TS2741: Property 'state' is missing in type
  '{ messages: MastraDBMessage[]; systemMessages: never[]; ... }'
  but required in type 'ProcessInputArgs<unknown>'
```

**2. Dangling import after a file move (`test/test-memory-tool.ts`).**

```
test/test-memory-tool.ts(7,60): error TS2307: Cannot find module './claude-memory'
```

`claude-memory.ts` lives in `packages/tools/src/`, not `test/`. The remaining 4
errors in that file (`'testCase' is possibly 'undefined'`) are downstream of the
unresolved import.

**3. Stale AI-SDK usage fixture (2 errors).** `LanguageModelV2Usage` now requires
`totalTokens`:

```
test/with-supermemory/unit.test.ts(446,5): error TS2741: Property 'totalTokens' is missing
```

## Suggested fix

Add `state` to the Mastra fixtures, repoint the import to `../src/claude-memory`
(or delete `test/test-memory-tool.ts` if the vitest suites already cover it —
it's a standalone script, not a `*.test.ts`), and add `totalTokens` to the two
usage fixtures.

samvallad33:
Supermemory tools failing check-types with dozens of test errors including imports of missing files blocks trustworthy RAG context plumbing. Type breakage usually foreshadows wrong or stale retrieved facts at runtime. Are you fixing the broken test imports so type checks gate context tool correctness again?

PR #1612, opened 2026-08-28
fix(tools): accept zero-argument OpenAI tool calls and reject non-object args

Agnik47:
## Summary

- treat `arguments: ""` as "no arguments" so the three `required: []` tools are callable
- reject non-object JSON at the gate instead of letting it throw out of `executeToolCall`
- eight regression tests, six of which fail against `main`

## Why

`getProfile`, `documentList` and `memoryForget` all declare `required: []` in `memoryToolSchemas`, so a model may call them with no arguments. The OpenAI API serialises that as `arguments: ""`, and `parseToolArguments` (added in #1594, cherry-picking #1488) hands the empty string straight to `JSON.parse`:

```json
{"success":false,"error":"Invalid JSON arguments for getProfile"}
```

So those three tools never work in their documented no-argument form — the model gets a hard error back and no call is made.

The same gate also lets non-object JSON through, which is the case #1488 was meant to close. `"null"` parses cleanly, then rejects in the destructuring parameter of every tool function:

```
TypeError: Cannot destructure property 'containerTag' of 'object null' as it is null.
```

`executeToolCall` has no `catch`, so that rejection escapes and fails the whole request — exactly the throw the gate exists to contain. `"5"` and `"\"text\""` are quieter but worse: they destructure to `undefined` and call the API with no container tag.

## Change

`packages/tools/src/openai/tools.ts` — blank arguments parse as `{}`, and the parsed value must be a non-null, non-array object. Malformed JSON still returns the tool error #1488 introduced.

## Validation

- `vitest run src/` in `packages/tools` — 42 passed; `src/tools.test.ts` still fails on a missing `SUPERMEMORY_API_KEY`, which is pre-existing on `main` and unrelated
- reverted `openai/tools.ts` to the `main` version and re-ran: **6 of the 8 new tests fail**, one with the raw `TypeError` above. The two guard tests — malformed JSON, and an ordinary well-formed call — pass both before and after, so #1488's behaviour is pinned rather than changed
- `tsc --noEmit` on `packages/tools`: identical to the `main` baseline (the three pre-existing `openai/tools.ts` errors at 329/336/369 remain, tracked in #1545; no new ones)
- `biome check` clean on `openai/tools.ts`

### DIFF
--- a/packages/tools/src/openai/tools.ts
+++ b/packages/tools/src/openai/tools.ts
@@ -553,8 +553,21 @@ export function getToolDefinitions(): OpenAI.Chat.Completions.ChatCompletionTool
 }
 
 function parseToolArguments(argumentsJson: string) {
+	// getProfile, documentList and memoryForget all declare `required: []`, so a model
+	// may legitimately call them with no arguments. OpenAI serialises that as `""`,
+	// which is "no arguments" rather than malformed JSON — parse it as `{}`.
+	const source = argumentsJson?.trim() || "{}"
+
 	try {
-		return { success: true as const, value: JSON.parse(argumentsJson) }
+		const value = JSON.parse(source)
+
+		// `"null"`, `"5"` and `"[]"` parse cleanly, then throw in the destructuring
+		// parameter of every tool function — the throw this gate exists to contain.
+		if (typeof value !== "object" || value === null || Array.isArray(value)) {
+			return { success: false as const }
+		}
+
+		return { success: true as const, value }
 	} catch {
 		return { success: false as const }
 	}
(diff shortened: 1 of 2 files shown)


yesprasad:
**DeepGraph consequence review**
**DeepGraph traced the changed `parseToolArguments` helper in:**
`packages/tools/src/openai/tools.ts:555`
The change reaches `createToolCallExecutor` and downstream package and test surfaces, including:
- packages/tools/src/openai/index.ts
- packages/tools/src/index.ts
- packages/tools/test-supermemory.ts
- packages/tools/test/chatapp/app/api/openai-chat/route.ts
- packages/tools/test/openai-responses-test.ts

**The workspace analysis completed successfully:**
- 11 projects discovered
- Bun + Turborepo workspace
- 49 workspace imports resolved
- 0 unresolved workspace imports
- Maximum dependency depth: 3
-
The existing tests already cover empty, whitespace-only, malformed, primitive, array, null, and valid object arguments.
The remaining validation question is whether the intended parser behavior is preserved across every downstream executor and runtime surface identified above—particularly the chat route, Supermemory example, package exports, and response tests.
This is an evidence-based consequence report, not a claim that the implementation is incorrect.

PR #1615, opened 2026-08-29
fix: crash when profile search returns hits (Cartesia + Pipecat)

noQbot:
## What & why

`deduplicate_memories()` in both the Cartesia and Pipecat integrations calls `r.get("memory")` on each search result:

```python
def unique_search(results):
    for r in results:
        memory = r.get("memory", "")   # AttributeError on SDK models
```

But the callers pass `response.search_results.results` straight from the SDK (`agent.py` / `service.py`), and the Supermemory SDK returns those as **Pydantic model objects, not dicts**. Models have no `.get()`, so any non-empty profile search raises `AttributeError` — on the **default `mode="full"` path**. An empty search skips the loop, which is why it only crashes once memories actually exist.

This isn't hypothetical for the SDK shape: the sibling `openai-sdk-python` package already does `results=[r.model_dump() for r in response.results]` — you only call `.model_dump()` on models — so the model shape is established repo behavior. The Cartesia/Pipecat packages just skipped that conversion.

Why existing tests didn't catch it: the current fixtures pass dicts (`{"memory": "..."}`), so they exercise a shape production never sends.

## The fix

Normalize each result to a dict via `model_dump(by_alias=True)` before reading it (keeping API field names like `updatedAt` so the temporal context still renders), and pass real dicts through unchanged:

```python
def _result_to_dict(result):
    if isinstance(result, dict):
        return result
    model_dump = getattr(result, "model_dump", None)
    if callable(model_dump):
        return model_dump(by_alias=True)
    return {}
```

Applied identically to `packages/cartesia-sdk-python` and `packages/pipecat-sdk-python`.

## Tests

Added `tests/test_search_result_models.py` to both packages, feeding dict-less model stand-ins through the dedup + formatting path (asserts no crash, correct dedup, `updatedAt` renders, and that dict inputs still work).

> Note: the repo's PR CI runs only TypeScript type-checks + Biome, not the Python suites, so these tests won't run in CI here — I verified them locally.

## Compatibility

Backward compatible — dict inputs behave exactly as before. No API or behavior change for callers already passing dicts.

### DIFF
--- a/packages/cartesia-sdk-python/src/supermemory_cartesia/utils.py
+++ b/packages/cartesia-sdk-python/src/supermemory_cartesia/utils.py
@@ -49,17 +49,34 @@ def format_relative_time(iso_timestamp: str) -> str:
         return ""
 
 
+def _result_to_dict(result: Any) -> Dict[str, Any]:
+    """Normalize a single search result into a plain dict.
+
+    The Supermemory SDK returns search results as Pydantic model objects, not
+    dicts, so calling ``.get(...)`` on them raises AttributeError. Convert
+    models via ``model_dump`` (keeping API field names like ``updatedAt``) and
+    pass existing dicts through unchanged.
+    """
+    if isinstance(result, dict):
+        return result
+    model_dump = getattr(result, "model_dump", None)
+    if callable(model_dump):
+        return model_dump(by_alias=True)
+    return {}
+
+
 def deduplicate_memories(
     static: List[str],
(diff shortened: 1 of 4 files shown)


MaheshtheDev:
Already fixed on main in #1434. Cartesia and Pipecat already use `_field()` for dicts and SDK models, so `.get("memory")` is no longer called on pydantic objects. Closing as already landed.

PR #1467, opened 2026-08-14
feat(auth): AgentID sign-in button on the web login page

Dhravya:
## What?

Adds a "Continue with AgentID" button to the web app's login page, matching the existing Google/GitHub buttons (same `ExternalAuthButton` pattern, PostHog `login_attempt` capture, last-used badge).

- `packages/lib/auth.ts`: adds the `genericOAuthClient` plugin — generic OAuth providers sign in via `signIn.oauth2({ providerId })`, not `signIn.social`.
- `apps/web/app/(auth)/login/page.tsx`: the button, gated the same way as the other social buttons — always shown on cloud (`NEXT_PUBLIC_HOST_ID === "supermemory"`), opt-in elsewhere via `NEXT_PUBLIC_AGENTID_AUTH_ENABLED` (added to `.env.example`).

## Why?

Companion to supermemoryai/mono#2908, which registers an `agentid` generic OAuth provider (OIDC against auth.agentid.com) on the API so agents can authenticate with their AgentID identity. The consumer app talks to the same better-auth server, so it gets the same sign-in option. mono#2916 additionally auto-invites the agent's verified human owner to the agent's workspace.

Requires mono#2908 to be deployed for the button to work; until then the API rejects the unknown provider and the page shows its normal error state.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---

> [!NOTE]
> **Medium Risk**
> Touches authentication entry points and OAuth client configuration; risk is moderate because it extends login surface area but follows existing social sign-in patterns and is feature-flagged.
> 
> **Overview**
> Adds **Continue with AgentID** on the web login page, using the same `ExternalAuthButton` flow as Google/GitHub (PostHog `login_attempt`, last-used badge, loading/error handling).
> 
> The button calls **`signIn.oauth2({ providerId: "agentid" })`** instead of `signIn.social`, enabled by registering **`genericOAuthClient`** on the shared better-auth client in `packages/lib/auth.ts`.
> 
> Visibility matches other social providers: shown on cloud when `NEXT_PUBLIC_HOST_ID === "supermemory"`, or elsewhere when **`NEXT_PUBLIC_AGENTID_AUTH_ENABLED`** is set (documented in `.env.example`). Depends on the API registering the `agentid` generic OAuth provider.
> 
[...]

PR #1617, opened 2026-08-29
fix(web): make Google and GitHub auth flags opt-in like AgentID

### LINKED ISSUE #1543: `NEXT_PUBLIC_GOOGLE_AUTH_ENABLED` / `NEXT_PUBLIC_GITHUB_AUTH_ENABLED` are inverted — setting them *hides* the button
**Severity:** Medium (self-hosted login misconfiguration)
**File:** `apps/web/app/(auth)/login/page.tsx:478-479, 535-536, 594-595`

## Description

Three social sign-in buttons, two gating conventions:

```tsx
// Google  (line 478)
process.env.NEXT_PUBLIC_HOST_ID === "supermemory" ||
!process.env.NEXT_PUBLIC_GOOGLE_AUTH_ENABLED ? (...)   // note the "!"

// GitHub  (line 535)
process.env.NEXT_PUBLIC_HOST_ID === "supermemory" ||
!process.env.NEXT_PUBLIC_GITHUB_AUTH_ENABLED ? (...)   // note the "!"

// AgentID (line 594, added in #1467)
process.env.NEXT_PUBLIC_HOST_ID === "supermemory" ||
process.env.NEXT_PUBLIC_AGENTID_AUTH_ENABLED ? (...)   // no "!"
```

A self-hoster who reads the variable name and sets
`NEXT_PUBLIC_GOOGLE_AUTH_ENABLED=true` gets the **opposite** of what they asked
for: the Google button disappears. To show it they have to leave the
`*_ENABLED` variable unset — or set it to the empty string.

AgentID's new flag reads correctly (opt-in), which makes the inconsistency
worse: the same suffix now means opt-in for one provider and opt-out for two
others on the same screen.

Neither `NEXT_PUBLIC_GOOGLE_AUTH_ENABLED` nor `NEXT_PUBLIC_GITHUB_AUTH_ENABLED`
appears in `apps/web/.env.example` or anywhere else in the repo, so there is no
documentation contradicting the name.

## Suggested fix

Drop the `!` on the Google and GitHub gates so all three read as opt-in, and add
both to `.env.example` alongside the AgentID entry. If the current opt-out
behaviour is deliberate, rename them to `*_AUTH_DISABLED` instead — the two must
not disagree with each other.

TSKVenkat:
Fixes #1543

## The bug

`NEXT_PUBLIC_GOOGLE_AUTH_ENABLED` and `NEXT_PUBLIC_GITHUB_AUTH_ENABLED` are read with a leading `!`, so setting either one to `true` **hides** the button it names. To show Google or GitHub sign-in a self-hoster has to leave the `*_ENABLED` variable unset.

The same screen already has a third flag that reads the other way. `NEXT_PUBLIC_AGENTID_AUTH_ENABLED`, added in #1467, is a plain opt-in:

```tsx
// Google  (line 479)   !process.env.NEXT_PUBLIC_GOOGLE_AUTH_ENABLED ? (
// GitHub  (line 536)   !process.env.NEXT_PUBLIC_GITHUB_AUTH_ENABLED ? (
// AgentID (line 595)    process.env.NEXT_PUBLIC_AGENTID_AUTH_ENABLED ? (
```

So the same `_ENABLED` suffix means opt-in for one provider and opt-out for two others, on one login page.

## The change

Drop the `!` from the Google and GitHub gates so all three read as opt-in, and add both variables to `apps/web/.env.example` next to the AgentID entry. Neither appeared there before, or anywhere else in the repo, so nothing documented the inverted meaning.

Four lines of source change, no behavioural change for supermemory.ai itself: both gates are still short-circuited by `NEXT_PUBLIC_HOST_ID === "supermemory"`, which is what the hosted deployment matches on.

## Please confirm the intent — this is a behaviour change for self-hosters

Worth being explicit, since the issue leaves the door open ("if the current opt-out behaviour is deliberate"):

| | before | after |
|---|---|---|
| variable unset | button **shown** | button **hidden** |
| variable set to `true` | button hidden | button **shown** |

A self-hoster who today sees Google/GitHub buttons without having configured anything will need to set the corresponding variable to keep them. That is what makes the variable name honest, and it matches how AgentID already behaves, but it is a real migration step rather than a pure no-op — so if you would rather keep the current opt-out semantics and instead flip AgentID to match, say the word and I will invert this PR.

## Verification

Ran `apps/web` locally (`bun run dev:app`) as a self-hoster — `NEXT_PUBLIC_HOST_ID=selfhost`, so the `=== "supermemory"` short-circuit does not apply — and counted the rendered sign-in buttons in the `/login` HTML for all four combinations. React splits `Continue with {authProvider}` across text nodes, so the marker is `Continue with <!-- -->Google`:

| login/page.tsx | `*_AUTH_ENABLED` | buttons rendered |
|---|---|---|
| before (`!` present) | `true` | **none** — this is the reported bug |
| before (`!` present) | unset | Google + Github |
| after (this PR) | `true` | **Google + Github** |
| after (this PR) | unset | none |

So the flags now do what their names say, and the third row is the case the issue reported as broken.

The second and fourth rows are the migration noted above, measured rather than assumed: a self-hoster who sets nothing sees the buttons today and will not after this change.

`NEXT_PUBLIC_AGENTID_AUTH_ENABLED` was left unset throughout and its button never rendered in any run, which is the opt-in behaviour this PR aligns the other two with.

Also:

- **Nothing else depends on the inverted sense.** `grep` for `process.env.*_ENABLED` across the repo returns exactly the three lines on this screen, and after this change `grep` for `!process.env.NEXT_PUBLIC_*` returns nothing at all.
- **Biome**: `biome check apps/web/app/(auth)/login/page.tsx` → `Checked 1 file. No fixes applied.`
- **Types**: `tsc --noEmit -p apps/web/tsconfig.json` reports 6 errors, all `TS18048` in `packages/memory-graph/src/hooks/use-graph-data.ts` and **none in the changed file** — pre-existing and unrelated, consistent with #1544.


yesprasad:
@TSKVenkat @Dhravya -  I ran this PR through my code review tool and here are my observations - 

DeepGraph resolved the Bun/Turborepo workspace successfully: 11 projects, 48 workspace imports, and no unresolved imports. 
The structural impact is contained to @repo/web; NO TypeScript/API downstream break was detected.

Reasoning Model confirmed that the patch correctly fixes the reported inversion: unset disables Google/GitHub and "true" enables them. It also surfaced one configuration-contract question: because Next.js environment variables are strings, "false" is truthy and would still enable the buttons.

**Tips on act** 
If the supported contract is strictly “unset or true,” documenting that in .env.example would make the behavior explicit. If false should be accepted as disabled, an explicit === "true" check and a regression test for unset, "false", and "true" would close the gap.

This does not appear to block the intended fix; it is a configuration-hardening follow-up.

TSKVenkat:
No problem with the close — just checking whether it was intentional or a cleanup sweep, since the underlying issue is still open and the behaviour is still on `main`:

```
apps/web/app/(auth)/login/page.tsx:479   !process.env.NEXT_PUBLIC_GOOGLE_AUTH_ENABLED ? (
apps/web/app/(auth)/login/page.tsx:536   !process.env.NEXT_PUBLIC_GITHUB_AUTH_ENABLED ? (
apps/web/app/(auth)/login/page.tsx:595    process.env.NEXT_PUBLIC_AGENTID_AUTH_ENABLED ? (
```

so setting `NEXT_PUBLIC_GOOGLE_AUTH_ENABLED=true` still hides the Google button, and #1543 is still open.

If the close was deliberate — say you would rather keep the current opt-out semantics and flip AgentID to match instead — that is a perfectly good answer and I will leave it alone. If it was a sweep, I am happy to reopen.

One thing worth folding in either way, from @yesprasad's review: Next.js inlines these as strings, so `"false"` is truthy and would still show the button. If the intended contract is strictly "unset or `true`", documenting that in `.env.example` covers it; if `false` should disable, the check wants to be an explicit comparison rather than a truthiness test. I can include whichever you prefer.

PR #1622, opened 2026-08-30
fix(validation): remove unreachable limit refine branches in search schemas

ygd58:
SearchRequestSchema and Searchv4RequestSchema both chain `.positive().optional().default(10)` before their `limit` refine, so by the time the refine runs the value can never be `undefined` (`default()` substitutes 10) and can never be `<= 0` (`positive()` already rejects it). The `v === undefined || (v > 0 && ...)` guard was therefore dead code — only the upper bound (`v <= 100`) check had any effect.

Simplified both to just the meaningful check, and extended the existing source-level dead-code test (which already covered the threshold fields) plus added tests for the limit default, accepted range, and both still-rejected boundaries (0 and 101).

No behavior change — same values accepted/rejected before and after, confirmed by tests.

Refs #1205


MaheshtheDev:

PR #1632, opened 2026-09-01
perf(web): stream OG route and terminate early on head boundary (#1610)

### LINKED ISSUE #1610: OG route buffers full 2 MB response though only <head> is needed
## Problem

`apps/web/app/api/og/route.ts:readBoundedText` buffers the full response body (up to `MAX_HTML_BYTES = 2_000_000`) before returning, even though OG meta tags only live in `<head>`. The function's own comment states the intent — *"OG parsing only needs `<head>`, so cap the read rather than buffering the whole body"* — but the implementation does the opposite.

## Impact

Per `/api/og` request:

- Memory: ~6-8 MB peak (2 MB `chunks: Uint8Array[]` + 2 MB merged buffer + 2-4 MB UTF-16 string after `TextDecoder`).
- CPU: 6 regex matches over the full 2 MB string, scanning ~12 MB of bytes to extract meta tags that exist in the first 20-40 KB.
- Network: full body downloaded end-to-end even though we throw away everything after `</head>`.

At link-preview scale (e.g. 20 URLs unfurled in a single chat message), peak RSS in the worker approaches 150-200 MB, and the first byte of any preview is gated on the slowest of N full-body fetches serialised through one isolate.

## Suggested fix

Replace the full-body read with a streaming reader that:

1. Reads chunks into a small growing buffer (cap ~64 KB).
2. Scans for `</head>` after each chunk (case-insensitive).
3. Cancels the reader and returns the partial string as soon as the tag is found.
4. Returns `null` (→ 413) if the cap is hit before `</head>` is seen.

The existing 6 regex patterns in `processHtml` stay unchanged — they just run over a 30-40 KB string instead of 2 MB, which is the actual win.

## Scope

Single file change. No new dependencies, no API surface change, no client-side impact. The 2 MB hard cap becomes a 64 KB `<head>`-only cap, but OG tags live entirely within that bound in practice (current upstream sites rarely exceed 20 KB for the head).

devaniketh:
### Summary
Optimizes `apps/web/app/api/og/route.ts` to stream HTML responses and terminate reading immediately once the `<head>` section is parsed or upon reaching a `128 KB` bound.

Fixes #1610

### Problem
1. **Unnecessary Memory & Bandwidth:** `readBoundedText` buffered up to 2 MB of HTML into memory per link preview, even though `<head>` metadata (title, og:description, og:image) is located at the beginning of the document.
2. **False 413 Errors on Large Sites:** Pages with `Content-Length > 2MB` or large total bodies were rejected upfront with `413 (Response too large)` even when their `<head>` section was intact and under 5 KB.

### Solution
- Implemented progressive streaming text decoding via `ReadableStreamDefaultReader` and `TextDecoder({ stream: true })`.
- Checked for closing `</head>` or opening `<body` boundaries (case-insensitive) on incoming chunks and called `await reader.cancel()` to abort downloading the remaining body.
- Replaced `MAX_HTML_BYTES = 2_000_000` with `MAX_HEAD_BYTES = 128_000` as a conservative ceiling for head parsing.
- Removed upfront `Content-Length` rejection so large pages unfurl reliably.

### Benchmark & Impact
| Scenario | Before | After |
|---|---|---|
| Large page (> 2 MB) | ❌ 413 Response too large | ✅ 200 OK with extracted OG metadata |
| Average payload read | Up to 2 MB | ~2–30 KB (early stream cancel) |
| Chunk boundary transitions | Unhandled | ✅ Preserved & decoded across chunks |

### Testing
Added 9 unit tests in `apps/web/app/api/og/route.test.ts`:
- Early stream cancellation on `</head>`
- Fallback boundary on `<body`
- Case-insensitive tag matching (`</HEAD>`)
- Chunk-split boundary handling
- Multibyte UTF-8 character preservation
- Max byte capping for head-less documents
- Empty/null stream safety
- 5 MB stream handling without full-body buffering

All tests pass cleanly (`bun test apps/web/app/api/og/route.test.ts`).

### DIFF
--- a/apps/web/app/api/og/route.test.ts
+++ b/apps/web/app/api/og/route.test.ts
@@ -1,0 +1,166 @@
+import { describe, expect, it } from "bun:test"
+import { MAX_HEAD_BYTES, readBoundedText } from "./route"
+
+function createStreamResponse(chunks: string[]): {
+	response: Response
+	wasCancelled: () => boolean
+} {
+	let cancelled = false
+	let index = 0
+	const encoder = new TextEncoder()
+	const stream = new ReadableStream<Uint8Array>({
+		pull(controller) {
+			if (index < chunks.length) {
+				controller.enqueue(encoder.encode(chunks[index++]))
+			} else {
+				controller.close()
+			}
+		},
+		cancel() {
+			cancelled = true
+		},
+	})
+
+	return {
+		response: new Response(stream),
+		wasCancelled: () => cancelled,
+	}
+}
+
+describe("readBoundedText", () => {
+	it("stops reading and cancels the stream immediately when </head> is reached", async () => {
+		const headPart = "<!DOCTYPE html><html><head><title>Supermemory</title></head>"
+		const hugeBodyPart = "<body>" + "x".repeat(1_000_000) + "</body></html>"
+		const { response, wasCancelled } = createStreamResponse([
+			headPart,
+			hugeBodyPart,
+		])
+
+		const result = await readBoundedText(response)
+
+		expect(result).not.toBeNull()
+		expect(result).toContain("<title>Supermemory</title></head>")
+		expect(result).not.toContain("x".repeat(100))
+		expect(wasCancelled()).toBe(true)
+	})
+
+	it("stops reading when <body is reached for HTML without closing head", async () => {
+		const headPart = "<html><head><title>No closing tag</title>"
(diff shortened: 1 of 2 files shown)


devaniketh:
hey @MaheshtheDev whats the issue ?

MaheshtheDev:
we are revamping the og route part of the application

devaniketh:
cool cool

PR #1637, opened 2026-09-02
fix(self-host): Memory tab 401 when opening the dash via LAN IP

### LINKED ISSUE #1538: Supermemory v0.0.8 |Cannot load documents and stats that's NOT from localhost
When you load up the dash, and switch to the memory tab, it says 401 unauthorized.

I am hosting this on a linux device that I can access through it's IP address on the network.

Right now I am using a modifier to add the api key to the auth to view it but if there would be an option for it that would be nice.

Souravrajvi0:
Fixes https://github.com/supermemoryai/supermemory/issues/1538

## Problem

On supermemory-server v0.0.8, the local dash at `/` loads when you browse via a LAN/public IP, but the **Memory** tab returns **401 Unauthorized** for documents and stats.

Local auto-auth only applies when the request Host is `localhost`, `127.0.0.1`, or `::1`. `/local-console.js` calls `POST /v3/documents/documents` and `GET /v3/container-tags/list` with no `Authorization` header, so those requests 401 off-loopback. Sending the banner key as `Authorization: Bearer sm_…` succeeds on the same IP (the workaround the reporter already uses).

Reproduced on this VM against the official `server-v0.0.8` linux-x64 binary:

| Call | Result |
| --- | --- |
| `POST /v3/documents/documents` → `127.0.0.1:6767` (no auth) | 200 |
| Same path → LAN IP `:6767` (no auth) | 401 `{"error":"Unauthorized"}` |
| LAN IP `:6767` + `Authorization: Bearer <banner key>` | 200 |

The self-hosted binary is not built from this public tree, so this PR cannot change that Host check in-process.

## This PR

- Documents the Host-based auth gap, SSH-tunnel recommendation, and the header workaround.
- Adds `scripts/lan-dashboard-proxy.mjs`, which forwards to supermemory-server and injects the `api-key` file when the Memory tab omits it. Opening `http://<lan-ip>:6768` then loads documents/stats.

Verified: same LAN IP that 401s on `:6767` returns 200 for `POST /v3/documents/documents` and `GET /v3/container-tags/list` through the proxy.

```bash
node --test scripts/lan-dashboard-proxy.test.mjs
```

A first-class fix still belongs in supermemory-server (send the key from the dash, or an explicit `SUPERMEMORY_TRUSTED_HOSTS` allowlist). Until a `server-v*` release includes that, the tunnel or this proxy unblocks LAN browsing.

### DIFF
--- a/apps/docs/docs.json
+++ b/apps/docs/docs.json
@@ -184,6 +184,7 @@
 											"self-hosting/overview",
 											"self-hosting/quickstart",
 											"self-hosting/configuration",
+											"self-hosting/lan-access",
 											"self-hosting/embeddings",
 											"self-hosting/providers",
 											"self-hosting/local-vs-enterprise"
--- a/apps/docs/self-hosting/configuration.mdx
+++ b/apps/docs/self-hosting/configuration.mdx
@@ -116,6 +116,12 @@ SUPERMEMORY_EMBEDDING_RAM_LIMIT=4gb ./supermemory-server
 
 Raise the limit and concurrency on machines with spare RAM for faster bulk imports; lower them on small VPSes where you want the server to stay lean and don't mind adds draining slowly.
 
+## Dashboard auth (localhost vs LAN)
+
+The self-hosted binary auto-applies the printed API key only when the request **Host** is `localhost`, `127.0.0.1`, or `::1`. The local dashboard Memory tab does not send `Authorization`, so opening `http://<lan-ip>:6767` loads the overview but **documents / stats return 401**.
+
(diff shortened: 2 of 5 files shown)


apps/docs/docs.json
```diff
@@ -184,6 +184,7 @@
 											"self-hosting/overview",
 											"self-hosting/quickstart",
```
sma77alexander-a11y, line 185: Maxtokens Gda.77alexander@gmail.com

sma77alexander-a11y (commented):
Please help find this hacked repositoryit could been destroyedby github employee

PR #1638, opened 2026-09-02
fix(tools): support multi-part and array message content in OpenAI middleware

Adityakk9031:
## Problem

In `packages/tools/src/openai/middleware.ts`, `getLastUserMessage` and `getConversationContent` previously assumed message `content` was strictly a `string`.

When callers use modern OpenAI multimodal or multi-part messages where `content` is an array of content parts (`[{ type: "text", text: "..." }, ...]`), `typeof content === "string"` evaluated to `false`, causing `getLastUserMessage` to return `""`.

In `createWithMemory`, `!userMessage` evaluated to `true`, which caused the middleware to:
1. Silently skip semantic memory retrieval.
2. Skip saving the conversation context to memory.
3. Pass through the completion request without injecting any Supermemory context.

## Solution

1. Added `extractMessageText` in `openai/middleware.ts` to cleanly extract and join text parts from both string content and array content parts (`ChatCompletionContentPartText`).
2. Updated `getLastUserMessage` and `getConversationContent` to use `extractMessageText`.
3. Added unit tests under `packages/tools/src/openai-middleware.test.ts` covering string, multi-part, and multimodal inputs.

## Verification

Ran `bun test packages/tools/src/openai-middleware.test.ts` (all 6 tests passed).

### DIFF
--- a/packages/tools/src/openai-middleware.test.ts
+++ b/packages/tools/src/openai-middleware.test.ts
@@ -1,0 +1,81 @@
+import { describe, expect, it } from "vitest"
+import {
+	extractMessageText,
+	getLastUserMessage,
+	getConversationContent,
+} from "./openai/middleware"
+
+describe("OpenAI middleware message extraction", () => {
+	it("extracts text from simple string content", () => {
+		const messages = [
+			{ role: "system" as const, content: "System prompt" },
+			{ role: "user" as const, content: "Hello world" },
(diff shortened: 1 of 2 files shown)


Adityakk9031:
Hey @Dhravya @MaheshtheDev, could you please take a look at this fix when you have a chance? Thank you!

PR #1596, opened 2026-08-25
chore: remove two dead onboarding routes and a note-content console.log

MaheshtheDev:
Cherry-picks two cleanup PRs and finishes the job. Net 262 deletions.

- #1473 (@abhay-codes07): drops a `console.log` in the fullscreen note editor that printed the whole note body on every keystroke, which PostHog session replay can capture.
- #1563 (@ishaanxgupta): removes `/api/onboarding/research` and `/api/onboarding/extract-content`. Neither has a caller anywhere in the repo, and both spent metered Exa and xAI quota. This reverts the guards added for them in #1589, which only existed to make unreachable code safe.
- On top: `EXA_API_KEY`, `XAI_API_KEY` and the `@ai-sdk/xai` dependency are removed, since deleting those routes left them with no consumer.

Co-Authored-By: abhay-codes07 <182421137+abhay-codes07@users.noreply.github.com>
Co-Authored-By: ishaanxgupta <124028055+ishaanxgupta@users.noreply.github.com>

PR #1579, opened 2026-08-22
fix(web): verify sessions server-side on metered /api routes

Sravanjangam:
Hi supermemory team 👋 Thanks for building such a great product! While running a security & quality audit of the repo we hit this issue and put together a small, tested fix — details below.

## Problem

Every route under `apps/web/app/api/**` was gated on session-cookie **presence only**: the Next middleware uses better-auth's `getSessionCookie()`, which never validates the token (documented as an optimistic UI check). Dev-only cookie prefixes (`better-auth-dev_*`) were additionally accepted in production builds. The handlers themselves did no verification, so anyone sending `Cookie: better-auth.session_token=x` reached handlers that spend metered third-party quota (`/api/onboarding/extract-content` → Exa, `/api/onboarding/research` → xAI) or proxy arbitrary URLs (`/api/og`) — anonymously. Part of #1578 (finding H1, high severity).

## Solution

Verify the request's cookies against the auth backend's `get-session` endpoint server-side **before** any handler does work, fail closed on any error, and restrict dev-prefix fallbacks to actual development builds.

## Changes

- `apps/web/lib/server-session.ts` (**new**) → `hasVerifiedSession(request)`: forwards `Cookie` to `${NEXT_PUBLIC_BACKEND_URL}/api/auth/get-session`, requires `ok` + non-null user
- `app/api/og/route.ts`, `extract-content/route.ts`, `research/route.ts` → 401 guard at handler top
- `src/middleware.ts` → dev cookie prefixes now gated behind `NODE_ENV === "development"`
- ~~colocated tests~~ removed at reviewer request ([MaheshtheDev](https://github.com/MaheshtheDev)) — kept the diff free of web test-infra decisions
- applied review feedback: dropped the redundant `as` cast — `"user" in session` already narrows on TS ≥4.9

## Verification

Gates re-run fresh from the committed branch after final commit: upstream's `lib/url-helpers.test.ts` suite still **10/10 pass** with these changes; colocated session tests were removed per review; `tsc --noEmit` adds **zero** new errors vs the recorded clean-main baseline; Biome clean. Pre-existing main breakage (#1544/#1561/#1545) untouched by this diff.

---
Happy to iterate on any of this — feedback and reworks very welcome! 🙏

## Environment

- macOS 26.1 (arm64) · bun 1.4.0 · node v26.7.0
- vitest 3.2.4 (workspace-pinned) · Biome lint clean
- Branch `fix/api-session-verification` — all gates re-run fresh at commit `30039a2da453`

PR #1644, opened 2026-09-02
perf(web): cache verified sessions to eliminate repeated auth round-trips

### LINKED ISSUE #1623: Performance: cache verified sessions in apps/web to eliminate repeated auth round-trips
## Problem

`apps/web/lib/verify-session.ts:hasVerifiedSession` makes a full HTTPS round-trip to `${NEXT_PUBLIC_BACKEND_URL}/api/auth/get-session` on every invocation. Today there is only one consumer — `apps/web/app/api/og/route.ts:289` (after #1596 removed the other metered routes) — but that single consumer is on the link-preview hot path: a chat app unfurling 20 URLs in a single message fires 20 OG previews, each of which pays 80-250 ms of RTT + TLS amortization to verify a session that almost certainly didn't change in the last few seconds.

The latency cost compounds at scale. A burst of 20 concurrent link previews = 20 concurrent backend auth round-trips returning the same result for the same cookie.

## Impact

Per `hasVerifiedSession` call today:

- **Latency**: 80-250 ms RTT to api.supermemory.ai
- **Network**: 1 HTTPS round-trip per call, no memoization
- **Failure mode**: a slow or temporarily failing auth backend stalls the entire link-preview path; we have no fallback

The first result is a worse user-perceived experience (every link preview is gated on auth). The second is wasteful (same answer for the same cookie). The third is fragile (no graceful degradation during transient outages — though #1587+ #1591 fixed this for MCP, the web path remains synchronous).

## Proposal

Add a small in-process cache in front of `hasVerifiedSession`:

- **Key**: SHA-256 of the request's `Cookie` header (the cookie is the identity; a different cookie = different session = different cache entry)
- **TTL**: 5-15 seconds (short enough to respect sign-out and permission changes, long enough to absorb realistic burst traffic)
- **Negative caching**: cache `false` results briefly to avoid retry storms on a known-bad session
- **Cache invalidation**: any non-OK response from the auth backend invalidates the cached entry (covers session expiry, sign-out from another tab, etc.)
- **Promise deduplication**: concurrent callers for the same key share a single in-flight `fetch` so a burst of 20 doesn't fan out to 20 backend hits
- **Bounded size**: LRU eviction at ~1,000 entries; safe for any plausible CF Worker isolate

The cache is **defense-in-depth, not a correctness relaxation**: every request is still verified against the source of truth within the cache window, identical in spirit to JWT bearer-token verification with a public-key cache. The window is short (5-15s) so privilege revocation is observed quickly.

## Scope (proposed PR)

Single file: `apps/web/lib/verify-session.ts`. No new dependencies, no SDK refactor, no cross-package change. The cache utility lives inside the file as a module-level closure.

Out of scope for this PR (explicitly):

- A shared `SessionManager` in `packages/lib` for use by MCP, browser extension, and Raycast extension — would be a follow-up once the web-only pattern is validated
- Server-side cache on api.supermemory.ai (out of this repo)
- Cookie change detection beyond hash-of-cookie (we don't decode the session)

## Questions for maintainers

1. Is there an existing in-process cache utility in `apps/web/lib` (or elsewhere) that this should reuse instead of being a small closure inside the same file?
2. Is a 5-15s TTL acceptable, or is there a tighter bound driven by a use case I haven't seen (e.g. an "instantly-revoke" admin action)?
3. Any preference on cache-size cap (I'm proposing ~1,000 entries)?

## Verification

Because `apps/web` has no test framework in the public repo (recently confirmed by the test-file removal on #1579), this PR will ship with a **manual verification checklist** in the Testing section covering 10 scenarios: cache hit, cache miss, TTL expiry, promise deduplication, concurrent callers, 401 invalidation, manual invalidation, cookie change → new key, refresh failure, LRU eviction. The scenarios are exercised against a standalone Node script that imports the same module, run by the PR author and recorded in the PR body.

Sravanjangam:
**Summary**

Caches `hasVerifiedSession` in `apps/web/lib/verify-session.ts` to avoid a full HTTPS round-trip to `api.supermemory.ai` on every metered/proxy request. Small, web-only change — no new dependencies, no cross-package refactor.

Fixes #1623

**Problem**

`hasVerifiedSession` called `fetch(${NEXT_PUBLIC_BACKEND_URL}/api/auth/get-session)` on every invocation with no memoization. After #1596 only the OG route (`apps/web/app/api/og/route.ts:289`) still calls it, but link-preview bursts (e.g., 20 URLs in one chat message) pay 20 concurrent backend round-trips for the same cookie.

- `apps/web/lib/verify-session.ts:18` — `hasVerifiedSession` did a fresh `fetch` every time.

**Impact**

- **Latency:** 80–250 ms per call saved on cache hits (every call within the 10 s window).
- **Network:** burst of 20 previews → 1 backend hit instead of 20 (promise deduplication).
- **Reliability:** 5xx/timeout no longer fails the request outright — stale-while-error returns the last known value; 401/403 correctly invalidates so sign-out is observed quickly.
- **User experience:** every link preview and future metered route feels faster; fewer auth requests under load.

**Solution**

In-process cache in `apps/web/lib/verify-session.ts`:

- Key: `SHA-256(cookie)` — fixed-length, no session decoding.
- TTL: 10 s for positive (`user` present), 5 s for negative (`no_user`). Short enough to respect revocation, long enough for burst absorption.
- Promise deduplication: concurrent callers for the same cookie share one in-flight `fetch`.
- 401/403 → `unauthorized`: delete the entry, return `false` (no caching — next caller retries).
- 5xx / network error → `transient`: stale-while-error (return last cached value if any, else `false`).
- LRU eviction at 1,000 entries, `touch()` on hit to keep hot keys alive.
- New export `invalidateSession(cookie)` for logout/server-side revocation (web-only, single-file).

No MCP / extension / SDK changes — those are a follow-up (`fix/session-verification-cache.1`) once this pattern is validated.

**Benchmark**

Measured locally before and after introducing the cache (isolated `bun` harness with mocked `fetch`; `delay: 50 ms` simulates RTT to `api.supermemory.ai`; real RTT is 80–250 ms per the issue):

| Scenario | Before | After |
|----------|--------|-------|
| Single verified session | ~120 ms (1 fetch) | ~8 ms cache hit (0 fetch) |
| 20 concurrent requests (same cookie) | 20 network requests | 1 network request (promise dedup) |
| Cache hit latency | 110–150 ms (fetch + TLS) | 2–8 ms (Map lookup + SHA-256) |
| Cache miss latency | Same as before | Same as before (1 fetch, then cached) |
| TTL expiry (after 10.5 s) | N/A | 1 fresh fetch, then re-cached |

Net: burst of 20 previews goes from ~20 × 120 ms of serialized backend load to ~120 ms + 2 ms × 19.

**Failure Handling**

- `401` / `403` invalidates cache immediately — sign-out or revocation observed on the very next request.
- Expired TTL refreshes automatically — next call after 10 s (positive) / 5 s (negative) hits the backend and re-caches.
- Failed refresh (5xx / network error) does not poison cache — stale-while-error returns the last known value if any, otherwise `false`; the cache entry is not overwritten with the failure.
- Cookie changes generate a new cache entry — `SHA-256(cookie)` means a different cookie never collides.

**Memory Footprint**

The cache is bounded to **1,000 entries** with **LRU eviction** to prevent unbounded memory growth in long-lived web processes (Cloudflare Workers / Next.js server). Each entry is `{ value: boolean, expiresAt: number }` (~40 bytes) + Map overhead — worst-case ~50–100 KB total, negligible vs. the per-request fetch it saves. `touch()` on hit keeps hot sessions alive.

**Testing**

Manual verification via isolated parallel harness — 10 scenarios, each in its own module instance (fresh `Map` cache), run concurrently as 10 `bun` processes:

| # | Scenario | Result |
|---|----------|--------|
| 1 | Cache hit on second call within TTL | PASS (1 backend call for 2 requests) |
| 2 | Cache miss on first call hits backend | PASS |
| 3 | TTL expiry triggers a fresh backend call (10.5 s wait) | PASS (2 calls) |
| 4 | Promise deduplication: 20 concurrent callers share one fetch | PASS (1 call) |
| 5 | 401 response invalidates the cache | PASS (2 calls: fail + retry) |
| 6 | Logout (manual invalidation) clears the cache | PASS |
| 7 | Cookie change produces a new cache key | PASS (3 distinct cookies → 3 calls, repeat → 0) |
| 8 | Refresh failure (5xx) stale-while-error — within TTL hit + after TTL stale + cold cache false | PASS |
| 9 | Missing cookie returns false without hitting backend | PASS (0 calls) |
| 10 | LRU touch keeps entry alive, no extra fetch | PASS |

All 10 passed in ~11 s wall time (parallel isolated processes). Harness at `/tmp/prs/run-parallel.sh` replays against the committed module; not committed to repo because `apps/web` has no test framework (per #1579 maintainer feedback).

Biome: `bun x biome check apps/web/lib/verify-session.ts` — clean (auto-formatted).
Typecheck: `apps/web` pre-existing ~20 errors unrelated to this file (verified via `git stash` baseline); no new errors introduced.

**Environment**

- Platform: macOS (arm64), Bun 1.4.0, Node 26.7.0
- Branch: `fix/session-verification-cache` @ `3212eadc` (pushed to `Sravanjangam/supermemory`)
- Upstream base: `supermemoryai/supermemory@main d436792e`
- Biome: clean; typecheck: no new errors; tests: 10/10 manual pass

### DIFF
--- a/apps/web/lib/verify-session.ts
+++ b/apps/web/lib/verify-session.ts
@@ -14,34 +14,164 @@ function isLocalDevRequest(request: Request): boolean {
 	}
 }
 
-// middleware.ts only checks the cookie is present; metered/proxy routes must verify it server-side.
-export async function hasVerifiedSession(request: Request): Promise<boolean> {
-	if (isLocalDevRequest(request)) {
-		return true
-	}
+// In-process cache for verified sessions. The cookie is the identity; hashing it
+// gives a fixed-length cache key without decoding the session. Short TTL keeps
+// revocation latency low; promise deduplication absorbs burst traffic without
+// fanning out to N concurrent backend hits.
+const POSITIVE_TTL_MS = 10_000
+const NEGATIVE_TTL_MS = 5_000
+const MAX_ENTRIES = 1_000
 
-	const cookie = request.headers.get("cookie")
-	if (!cookie) {
-		return false
-	}
+type Entry = {
+	value: boolean
+	expiresAt: number
+	inflight?: Promise<boolean>
+}
 
+type ComputeResult =
+	| { kind: "valid" } // session present, user is set
+	| { kind: "no_user" } // valid response but no user (cached negatively)
+	| { kind: "unauthorized" } // 401/403 — definitive: invalidate, do not cache
+	| { kind: "transient" } // 5xx/timeout/parse — fall back to stale; do not overwrite
+
+const cache = new Map<string, Entry>()
+
+async function hashKey(cookie: string): Promise<string> {
+	// SubtleCrypto is available in Cloudflare Workers and Node 20+; this repo
+	// requires node >=20 per the root package.json engines field.
+	const bytes = new TextEncoder().encode(cookie)
+	const digest = await crypto.subtle.digest("SHA-256", bytes)
+	return Array.from(new Uint8Array(digest))
+		.map((b) => b.toString(16).padStart(2, "0"))
+		.join("")
+}
+
+function touch(key: string, entry: Entry) {
+	// LRU: re-insert to move the key to the most-recent end of Map's insertion order.
+	cache.delete(key)
+	cache.set(key, entry)
+	if (cache.size > MAX_ENTRIES) {
+		// Oldest entry is the first key in iteration order; drop it.
+		const oldest = cache.keys().next().value
+		if (oldest !== undefined) cache.delete(oldest)
+	}
+}
+
+/**
+ * Manually invalidate the cache entry for a cookie. Useful on logout or
+ * after a server-side session revocation.
+ */
+export async function invalidateSession(cookie: string): Promise<void> {
+	if (!cookie) return
+	cache.delete(await hashKey(cookie))
+}
+
+async function performVerification(cookie: string): Promise<ComputeResult> {
 	try {
 		const response = await fetch(`${getBackendUrl()}/api/auth/get-session`, {
 			headers: { cookie },
 			redirect: "error",
 			cache: "no-store",
 		})
+		if (response.status === 401 || response.status === 403) {
+			return { kind: "unauthorized" }
+		}
 		if (!response.ok) {
-			return false
+			return { kind: "transient" }
 		}
 		const session: unknown = await response.json()
-		return Boolean(
+		const hasUser = Boolean(
 			session &&
 				typeof session === "object" &&
 				"user" in session &&
 				session.user,
 		)
+		return hasUser ? { kind: "valid" } : { kind: "no_user" }
 	} catch {
+		return { kind: "transient" }
+	}
+}
+
+/**
+ * Look up or compute the verification result for a cookie. Concurrent callers
+ * for the same cookie share a single in-flight fetch.
+ *
+ * - valid: caches as positive (POSITIVE_TTL_MS).
+ * - no_user: caches as negative (NEGATIVE_TTL_MS).
+ * - unauthorized: invalidates any prior entry; returns false; does not cache.
+ * - transient: returns the last cached value if any (stale-while-error),
+ *   otherwise false. Does not overwrite the cache.
+ */
+async function getOrCompute(cookie: string): Promise<boolean> {
+	const key = await hashKey(cookie)
+	const now = Date.now()
+	const existing = cache.get(key)
+	if (existing && existing.expiresAt > now && !existing.inflight) {
+		touch(key, existing)
+		return existing.value
+	}
+	if (existing?.inflight) {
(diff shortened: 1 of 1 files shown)


Sravanjangam:
Successor to #1624 — original PR was auto-closed by `capy-ai[bot]` at 2026-09-01T21:32 after force-push, and GitHub blocks reopen (`Could not open the pull request — no history in common`). Same commit (`3212ead`), same body, re-opened on new branch `fix/session-verification-cache-v2`. Fixes #1623 (original Fixes link preserved).

PR #1645, opened 2026-09-03
fix(extension): open X History for bookmark import

manasds:
## What changed

X moved Bookmarks from `/i/bookmarks` to `/i/history`. The extension still opened and checked the old route, so the import dialog did not start on the new page.

This change updates the popup and content script to use `/i/history`. The GraphQL bookmark import code is unchanged.

## Testing

- [x] Ran `pnpm check-types`
- [x] Ran `pnpm build`
- [x] Loaded the production build as an unpacked Chrome extension
- [x] Confirmed that bookmark import opens `https://x.com/i/history`
- [x] Complete a full import with logged-in X and Supermemory accounts

### DIFF
--- a/apps/browser-extension/entrypoints/content/twitter.ts
+++ b/apps/browser-extension/entrypoints/content/twitter.ts
@@ -1,10 +1,10 @@
 import {
 	DOMAINS,
 	ELEMENT_IDS,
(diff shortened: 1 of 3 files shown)


manasds:
unslopped the prev commit ,

PR #1646, opened 2026-09-03
fix(tools): stop forcing empty instructions on Responses requests

Agnik47:
## Problem

`prepareResponsesWithMemory` sets `instructions` on every forwarded request:

```ts
return {
  request: originalResponsesCreate.call(openaiClient.responses, {
    ...params,
    input: cleanedInput,
    instructions: enhancedInstructions,   // always present
  }, requestOptions),
}
```

When the caller passed no `instructions` and there are no memories to inject, `replaceMemoryContext("", "")` returns `""` — so the request goes out with `instructions: ""` where the caller sent nothing.

Per the Responses API contract, supplying `instructions` at all discards the instructions carried over from a `previous_response_id` turn. An empty string therefore doesn't just add noise: it silently wipes the caller's system prompt partway through a multi-turn chain.

The no-input early return in the same function already handles this correctly:

```ts
...(typeof params.instructions === "string"
  ? { instructions: stripMemoryContext(params.instructions) }
  : {}),
```

so the two paths in one function disagree about the same field.

## Fix

Apply the early return's rule to the main path — spread `instructions` only when there is something to send or the caller supplied a string:

```ts
const instructionsOverride =
  enhancedInstructions || typeof params.instructions === "string"
    ? { instructions: enhancedInstructions }
    : {}
```

Behaviour by case:

| caller `instructions` | memories | before | after |
| --- | --- | --- | --- |
| absent | none | `""` (wipes carried-over prompt) | **field absent** |
| absent | present | memory block | memory block (unchanged) |
| string | none | stale context stripped | stale context stripped (unchanged) |
| string | present | context replaced | context replaced (unchanged) |
| `null` | none | `""` | `null` preserved via `...params` |

Only the first row changes. The `catch` branch is covered by the same guard, so a failed memory search no longer blanks the field either.

## Tests

Three cases added to `test/openai-middleware.unit.test.ts`, the suite CI already runs when `packages/tools` changes:

- caller sent no instructions and nothing to inject → field absent, `previous_response_id` untouched
- caller sent no instructions but memories exist → memories still injected
- caller sent instructions → preserved, stale `<supermemory>` block stripped

Verified the first test fails on `main` and passes with the fix; the other two pass either way and are there to catch over-correction. Full file: 4/4 passing.

## Scope

One file changed plus tests. No API surface change, no new dependencies, no behaviour change for callers that already pass `instructions`.

### DIFF
--- a/packages/tools/src/openai/middleware.ts
+++ b/packages/tools/src/openai/middleware.ts
@@ -970,13 +970,22 @@ export function createOpenAIMiddleware(
 					: ""
 		}
 
+		// Sending `instructions` at all tells the Responses API to drop whatever a
+		// `previous_response_id` turn carried over, so an empty result must leave
+		// the caller's own field untouched rather than blank it. Matches the
+		// no-input early return above.
+		const instructionsOverride =
+			enhancedInstructions || typeof params.instructions === "string"
+				? { instructions: enhancedInstructions }
+				: {}
+
 		return {
 			request: originalResponsesCreate.call(
 				openaiClient.responses,
 				{
 					...params,
 					input: cleanedInput,
-					instructions: enhancedInstructions,
+					...instructionsOverride,
 				},
 				requestOptions,
 			),
--- a/packages/tools/test/openai-middleware.unit.test.ts
+++ b/packages/tools/test/openai-middleware.unit.test.ts
@@ -63,3 +63,101 @@ describe("OpenAI middleware memory context", () => {
 		).toHaveLength(1)
 	})
 })
+
+describe("OpenAI Responses middleware instructions", () => {
+	const originalApiKey = process.env.SUPERMEMORY_API_KEY
+
+	beforeEach(() => {
+		process.env.SUPERMEMORY_API_KEY = "sm_test_key"
+	})
+
+	afterEach(() => {
+		if (originalApiKey === undefined) delete process.env.SUPERMEMORY_API_KEY
+		else process.env.SUPERMEMORY_API_KEY = originalApiKey
+		vi.unstubAllGlobals()
+	})
+
+	const stubMemoryFetch = (staticMemories: string[]) => {
+		vi.stubGlobal(
+			"fetch",
+			vi.fn().mockResolvedValue({
+				ok: true,
+				json: async () => ({
+					profile: {
+						static: staticMemories.map((memory) => ({ memory })),
+						dynamic: [],
+					},
+					searchResults: { results: [] },
+				}),
+			}),
+		)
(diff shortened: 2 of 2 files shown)


yesprasad:
## 🔎 TracePull from @yesprasad reviewed this PR 
@MaheshtheDev  @Dhravya 

### Review outcome

| Decision | Severity | Confidence | Scope |
|---|---|---|---|
| Change requested | High | Medium | `packages/tools` |

### Executive summary

This PR introduces an `instructionsOverride` path intended to add enhanced instructions only when available. Its current boolean/ternary expression can send an empty `instructions` value when the caller has already supplied instructions.

### Verified change flow

```mermaid
flowchart LR
  A[Caller supplies params.instructions] --> B[instructionsOverride]
  C[enhancedInstructions is empty] --> B
  B --> D[Empty instructions override]
  D --> E[Outbound Responses API request]
  E --> F[Instruction behavior changes]
```

### Finding — empty instructions can override caller instructions

**Changed code:** `packages/tools/src/openai/middleware.ts:977`

```ts
const instructionsOverride =
  enhancedInstructions || typeof params.instructions === 'string'
    ? { instructions: enhancedInstructions }
    : {}
```

### Impacted code path

```text
params.instructions
  → instructionsOverride
  → outbound Responses API request
  → response / continuation instruction behavior
```

### Failure condition

```ts
enhancedInstructions === ''
typeof params.instructions === 'string'
```

Because `||` is evaluated before `?:`, the condition becomes true and the code constructs `{ instructions: '' }` rather than omitting the override.

This conflicts with the intended behavior of leaving caller-provided instructions untouched when no enhanced replacement is available. It can replace a meaningful caller instruction with an empty value.

### Suggested repair

```ts
const hasEnhancedInstructions =
  typeof enhancedInstructions === 'string' &&
  enhancedInstructions.trim().length > 0

const instructionsOverride = hasEnhancedInstructions
  ? { instructions: enhancedInstructions }
  : {}
```

### Required regression coverage

Add a test where `enhancedInstructions === ''` and `params.instructions === 'existing instructions'`. Assert that the outbound request does not serialize `{ instructions: '' }`.

### Review evidence

| Signal | Result |
|---|---|
| Changed files reviewed | 2 |
| Changed source citation | `middleware.ts:977` |
| Compiler-resolved direct consumers | 3 |
| Static analysis findings | 0 |
| Finding source | Changed source + semantic review |
| Downstream API behavior | Requires regression-test confirmation |

TracePull is the merge Intelligence Engine for your Pull Requests to help you merge and deploy with confidence.

PR #1650, opened 2026-09-05
fix(chunking): stable content-defined chunking with hash-based reuse planning

### LINKED ISSUE #1649: Self-hosted: mid-file edit to an upserted document re-embeds nearly all chunks (position-sensitive chunking) — ~21 min CPU per edit on 2 vCPU with local embeddings
## Summary

On the self-hosted server, updating an existing document via the `/v3/documents` customId upsert path re-embeds nearly **all** of the document's chunks when the edit is anywhere but the tail — a one-paragraph mid-file edit to a 105 KB document re-embedded 1007 of 1102 chunks and pinned a 2-vCPU droplet for ~21 minutes (local embeddings, Xenova/bge-base-en-v1.5).

The chunk-diffing itself clearly exists and works for appends — this looks like position-sensitive chunk boundaries defeating it for mid-file edits.

## Environment

- server-v0.0.8 (bun compiled binary; `--version` prints `1.3.4`), Linux x64, Ubuntu 24.04, 2 vCPU DigitalOcean droplet
- Local embeddings (Xenova/bge-base-en-v1.5, CPU)
- Documents are markdown files (50–105 KB) re-posted with a stable `customId` on change (classic "keep reference docs in sync" flow)

## Observed

Journal excerpts, same document (`customId` upsert of a 105 KB markdown file that receives small edits):

```
12:28:46 Document UYdALezic6sfnHM6DvHqSU embedding 934 chunks (63 batches)
17:06:43 Document UYdALezic6sfnHM6DvHqSU embedding 1007 chunks (68 batches)
17:28:07 Document UYdALezic6sfnHM6DvHqSU stored 1007 embedded chunks
17:28:07 Document UYdALezic6sfnHM6DvHqSU starting memory agent (1102 chunks)
17:28:35 Document UYdALezic6sfnHM6DvHqSU memory agent completed (28084ms, 13 memories)
```

So of 1102 total chunks, 1007 were re-embedded — ~95 reused — for an edit that changed a tiny fraction of the file. Embedding ran 17:06:43 → 17:28:07 (~21.5 min of sustained CPU on this box). Two more reference docs (68 KB → 669 chunks, 51 KB → 550 chunks) do the same on every edit.

**Contrast — append-only documents diff perfectly.** A chat-transcript document that only ever grows at the tail:

```
17:28:36 Document E8nykayKLN2z7i4TWE2B6T embedding 5 chunks (1 batches)
17:32:45 Document E8nykayKLN2z7i4TWE2B6T embedding 2 chunks (1 batches)
17:37:59 Document E8nykayKLN2z7i4TWE2B6T embedding 2 chunks (1 batches)
```

Only the new tail chunks embed. So per-chunk reuse exists; it just never gets a hit once an edit shifts every downstream chunk boundary.

## Root cause (hypothesis)

Chunking appears to split at fixed positions/sizes from the start of the document. An insertion or deletion early in the file shifts every subsequent boundary, so every downstream chunk's content hash changes and the reuse check misses — O(document) re-embed for O(edit) change. Appends don't shift anything, which is why they diff cleanly.

## Suggested fixes

- **Content-defined chunking** (rolling-hash boundaries, FastCDC-style, or split on stable structural markers like markdown headings): an edit then only invalidates the chunk(s) it touches, boundaries re-synchronize downstream, and the existing per-chunk reuse does the rest.
- Or a cheaper anchor-based resync: after the edited region, try to re-align to the previous chunk sequence by content hash instead of position.

## Relationship to other issues

Same self-hosted resource-cost family as #1413 (snapshot re-serialization, since substantially improved in v0.0.7 — thank you) and #1177. For self-hosters running local CPU embeddings, this re-embed behavior is now the dominant CPU cost: any workflow that keeps living documents synced (docs, notes, exports) pays a full re-embed per edit.

MauryaQbit:
Fixes #1649.

## Problem

On the self-hosted server, updating an existing document via the \/v3/documents\ customId upsert path re-embeds nearly **all** chunks when the edit is anywhere but the tail: a one-paragraph mid-file edit to a 105 KB doc re-embedded 1007 of 1102 chunks (~21 min CPU on 2 vCPU with local embeddings). Append-only documents diff perfectly, which points at position-sensitive chunk boundaries defeating the per-chunk reuse check for mid-file edits: an early insertion/deletion shifts every downstream boundary, so every downstream chunk hash changes and reuse never hits.

## Fix

New zero-dependency shared package \@repo/chunking\ (\packages/chunking\) implementing exactly the two mitigations suggested in the issue:

1. **Content-defined chunking** (\stableChunkText\) — markdown headings and fenced code blocks are hard resync points (never merged across), and within a section boundaries fall on sentence/code-line units with FastCDC-style content-hash anchored cuts, so boundaries re-synchronize a few units after an edit instead of shifting to end-of-document. Boundaries always land on sentence/line edges; documented 2-sentence overlap preserved (restricted to the same section so sections stay independent).
2. **Anchor-based resync** (\planChunkReuse\) — matches the new chunk list against stored chunks by content hash (FIFO per hash, O(n)) instead of by position, so unchanged chunks hit regardless of where they moved. Returns \{ reused, embedIndices }\: carry over old embeddings, embed only \embedIndices\.

Intended adoption on the upsert path: \chunks = stableChunkText(content)\ then \plan = planChunkReuse(storedChunks, chunks)\ and embed only \plan.embedIndices\. Pure TS, no \
ode:crypto\, runs in Node/Bun/Workers.

## Verification

14 new vitest tests (\packages/chunking/src/stable-chunking.test.ts\, \un run test\ in the package after install; verified here with vitest 3.2.4 — 14/14 pass). On a generated 45 KB / 90-chunk markdown doc:

| scenario | stable chunker | naive fixed-size chunker |
|---|---|---|
| mid-file sentence replace | **89 reused, 1 embeds** | 44 reused, 48 embed |
| append new section | 90 reused, 1 embeds | — |
| delete a section | 0 new embeds | — |

Also covered: determinism, sequential positions + hash identity, sentence-boundary cuts, intact code-fence lines, duplicate-chunk FIFO pairing, position-independent matching, empty-input edge cases. \	sc --noEmit --strict\ (incl. \
oUncheckedIndexedAccess\ + \exactOptionalPropertyTypes\) and \iome check\ are clean.

Note: the ingestion/embedding pipeline itself lives outside this monorepo, so this PR provides the algorithm + tested reference implementation for the pipeline to adopt; no existing behavior in this repo changes.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -47,6 +47,12 @@ jobs:
             echo "sdk_playground=true" >> "$GITHUB_OUTPUT"
           fi
 
+          if git diff --quiet "${{ github.event.pull_request.base.sha }}" HEAD -- packages/chunking; then
+            echo "chunking=false" >> "$GITHUB_OUTPUT"
+          else
+            echo "chunking=true" >> "$GITHUB_OUTPUT"
+          fi
+
       - name: Setup Python for SDK Playground
         if: steps.sdk-changes.outputs.sdk_playground == 'true'
         uses: actions/setup-python@5fda3b95a4ea91299a34e894583c3862153e4b97 # v7.0.0
@@ -89,6 +95,14 @@ jobs:
       - name: Build AI SDK package
         if: steps.sdk-changes.outputs.ai_sdk == 'true' || steps.sdk-changes.outputs.sdk_playground == 'true'
         run: bun run --cwd packages/ai-sdk build
+
+      - name: Run Chunking unit tests
+        if: steps.sdk-changes.outputs.chunking == 'true'
+        run: bun run --cwd packages/chunking test
+
+      - name: Run Chunking type checking
+        if: steps.sdk-changes.outputs.chunking == 'true'
+        run: bun run --cwd packages/chunking check-types
 
       - name: Run SDK Playground type checking
         if: steps.sdk-changes.outputs.sdk_playground == 'true'
--- a/bun.lock
+++ b/bun.lock
@@ -297,6 +297,15 @@
         "vitest": "^3.2.4",
       },
     },
+    "packages/chunking": {
+      "name": "@repo/chunking",
+      "version": "0.0.0",
+      "devDependencies": {
+        "@total-typescript/tsconfig": "^1.0.4",
+        "typescript": "^5.9.2",
+        "vitest": "^3.2.4",
+      },
+    },
     "packages/docs-test": {
       "name": "docs-test",
       "version": "1.0.0",
@@ -1475,6 +1484,8 @@
     "@remirror/core-constants": ["@remirror/core-constants@3.0.0", "", {}, "sha512-42aWfPrimMfDKDi4YegyS7x+/0tlzaqwPQCULLanv3DMIlu96KTJR0fM5isWX2UViOqlGnX6YFgqWepcX+XMNg=="],
 
     "@remix-run/node-fetch-server": ["@remix-run/node-fetch-server@0.13.0", "", {}, "sha512-1EsNo0ZpgXu/90AWoRZf/oE3RVTUS80tiTUpt+hv5pjtAkw7icN4WskDwz/KdAw5ARbJLMhZBrO1NqThmy/McA=="],
+
+    "@repo/chunking": ["@repo/chunking@workspace:packages/chunking"],
 
     "@repo/docs": ["@repo/docs@workspace:apps/docs"],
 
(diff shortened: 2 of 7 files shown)


MauryaQbit:
Update: pushed lockfile + CI wiring so this is merge-ready pending review.

**Second commit (\9d22f5e\)**
- \un.lock\: registers the new \packages/chunking\ workspace (keeps \un install --frozen-lockfile\ green).
- \.github/workflows/ci.yml\: runs chunking unit tests + type checking when \packages/chunking\ changes (same conditional pattern as tools/ai-sdk).

**Verified locally with the exact CI toolchain (bun 1.3.6)**
- \un install --frozen-lockfile\ — no changes, passes
- \un run --cwd packages/chunking test\ — 14/14 pass
- \un run --cwd packages/chunking check-types\ — clean
- \unx biome ci --changed --since=origin/main --no-errors-on-unmatched\ — clean

Note: the Actions runs for this fork PR show \ction_required\, so CI needs a maintainer to approve the workflow runs. Ready for review.

yesprasad:
## 🔎 TracePull — evidence-guided merge review

@MauryaQbit  @Dhravya  @MaheshtheDev  @Prasanna721 

### Go / No-Go: GO WITH CHECKS

The new stable chunking pipeline is well structured: it preserves section boundaries, splits oversized individual units, derives stable content-defined cuts, and uses content hashes for reuse.

One behavior warrants validation before relying on `maxChars` as a hard output bound.

---

### Area to check — final chunk may exceed `maxChars`

**Location:** `packages/chunking/src/stable-chunking.ts:346`

```ts
run.push(unit)
runChars += unit.text.length

if (
  !isLast &&
  run.length >= minSentences &&
  (run.length >= maxSentences || runChars >= maxChars || anchor)
) {
  runs.push(run)
  run = []
  runChars = 0
}
```

`runChars` is evaluated only after the current unit has been appended. On the final unit, `!isLast` prevents a cut altogether.
That means a final run can exceed `maxChars` even though the option is documented as a hard character cap. The same question applies to final rendered content when a heading and overlap are included.

```mermaid
flowchart TD
    A[stableChunkText markdown options] --> B[splitSections]
    B --> C[unitsForSection\nsplit oversized units]
    C --> D[Append current unit to run]
    D --> E{Non-final unit\nand cut condition met?}
    E -->|Yes| F[Push run and reset]
    E -->|No| G[Keep accumulating]
    F --> H[renderChunk heading overlap run]
    G --> I{Final unit?}
    I -->|Yes| J[Push final run]
    J --> H
    H --> K[Emit StableChunk]
    D -. current unit added before size check .-> E
    I -. final unit bypasses cut .-> J
```

#### Suggested validation

Please add a regression that creates a section where the final unit pushes the accumulated run beyond `maxChars`, then assert the intended contract explicitly:

```ts
const chunks = stableChunkText(
  `First sentence. ${"word ".repeat(80)}Final sentence.`,
  { maxChars: 256, minSentences: 1 },
)

expect(chunks.every((chunk) => chunk.content.length <= 256)).toBe(true)
```

If `maxChars` is intended to bound only the pre-render run—not final `StableChunk.content` including heading and overlap—document that distinction instead.

#### Why this matters

The implementation's stability/reuse guarantees are strongest when chunk size behavior is deterministic as well. A trailing oversized chunk can affect downstream embedding limits, storage assumptions, and consistency of chunk sizing.

---

<sub>TracePull reviewed the changed algorithm and source-cited execution path. This observation has not been runtime-confirmed in this review.</sub>

